PR #4, opened 2023-02-10
Remove all healthcheck scripts

kajinamit:
We'll maintain these scripts in individual operators and inject the required files from operators instead of including them in container images.


rabi:
> We'll maintain these scripts in individual operators and inject the required files from operators instead of including them in container images.

What's the advantage of injecting these static scripts and making the operator do more work?  I can understand if it's from convenience pov and the intention is to move healthcheck to individual operators and have managed by individual teams.

kajinamit:
It's mainly from convinience PoV.

* healthcheck script is used by the health probe defined in pod definition so is more related to operators than actual software code. Distributing healthcheck scripts allow us to more easily update health probe as all its components are placed in a single repo.

* Also, this allows maintainer of individual operators be more responsible for healthchecks they implement

* Some operators such as cinder-operator have implemented its own healthcheck script already and we don't have to pull in that work into this repo, which makes the core definition more complicated.

I'll propose this in the next sync to hear feedback from the team and will remove WIP if we agree with the direction.

SeanMooney:
for nova and placement please do remove the health check we want to fully control that from our operators

the current nova-compute checks for example are not something we want to support as it not really a correct health check
https://github.com/openstack-k8s-operators/tcib/blob/main/healthcheck/nova-compute

kajinamit:
no objections for 2 weeks. good to merge !

fao89 (approved):

bshephar (approved):

PR #5, opened 2023-02-10
Get rid of external images

kajinamit:
These references to external images are not really used and can be dropped. This allows us to remove the logic to detect image sources.


rabi (approved):
/lgtm

PR #6, opened 2023-02-14
Add a dev mode source file for local development

steveb:
This copies the tripleo-ansible/ansible-test-env.rc approach of allowing image definition files to come from the local git checkout instead of an installed location.


steveb:
What do you think of this approach, we could elaborate over time.

Also, a .gitignore would be nice :)

rabi (approved):

PR #8, opened 2023-02-20
Remove tcib rpm package

raukadah:
Earlier openstack-tripleo-common-containers rpm package was used to provide container-images kolla files.
Since due to tripleo component depreaction,
this package will not longer exists.

It got replaced with tcib, which also does not
exists and it breaks the image build process.

While building image one can pass
```
--volume /usr/share/tcib/container-images:/usr/share/tcib/container-images:z
```
to fix the issue.

### DIFF
--- a/container-images/tcib/base/base.yaml
+++ b/container-images/tcib/base/base.yaml
@@ -71,7 +71,6 @@ tcib_packages:
   - socat
   - sudo
   - tar
-  - tcib
   - util-linux-user
   - which
   modules:


rabi:
@raukadah Please check with RedDel folks, as they don't want any changes to their current process.

yazug:
So this was being installed so that a few files could be copied into the right place in the images.  Healthcheck files and some of the startup scripts look for related RUN cp /usr/lib/tripleo-common/  type commands that are in the tcib yaml files.  It would be preferred from a downstream point of view to not use tcib-gather-files functionality as it causes extra work when prepping and building images downstream.

rabi:
I think tcib-gather-files  won't be  used if we use volume mounts. Would that be acceptable? Though once we build the tcib package we can install it like before.

yazug:
assuming volume mounts will break downstream builds as well though.

rabi:
We decided not to do this.

PR #10, opened 2023-02-23
Import modify image role in tcib

raukadah:
It will be used with openstack tcib container image hotfix command.

Signed-off-by: Chandan Kumar <raukadah@gmail.com>


rabi (approved):
/lgtm.

Though not clear what the hotfix process would be in the future, this belongs here atm.

PR #14, opened 2023-02-27
Install additional clients

kajinamit:
This ensures we install more client libraries so that we can use subcommands for all services supported by oko.

Note this change does not cover legacy clients because these do not work with clouds.yaml .


raukadah (approved):

rabi (approved):
Not sure how many would use the openstackclient pod. I am sure we would also ship the client packages for users to install and anywhere along with the client plugins they use. Please note that one broken client plugin can impact OSC.

PR #15, opened 2023-02-27
Update ci roles to work with build jobs

rlandy:
The following changes were needed to get
container build jobs to pass:

 - use openstack tcib container build command

### DIFF
--- a/ci/roles/build_containers/defaults/main.yaml
+++ b/ci/roles/build_containers/defaults/main.yaml
@@ -68,3 +68,4 @@ openstack_release: master
 
 # Roles related vars
 tcib_repo: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/tcib"
+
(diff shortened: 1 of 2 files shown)


ci/roles/build_containers/defaults/main.yaml
```diff
@@ -39,7 +39,7 @@ release: 9
 image_tag: ''
 
 # Image Prefix
-container_name_prefix: openstack
+container_name_prefix: podified
```
rabi, line 42: Why use podified? I would suggest we keep openstack. I don't like the term podified anyway.

container-images/tcib/base/base.yaml
```diff
@@ -71,7 +71,6 @@ tcib_packages:
   - socat
   - sudo
   - tar
-  - tcib
```
rabi, line 74: We can't remove this unless we use host files (either by volume mounts or tcib_gather_files) which is not what RelDel wants. We should package tcib first before testing it in CI I guess.

raukadah (approved):
Thanks Ronelle, it will unblock the container build job . We can follow up on container name prefix and tcib package (https://review.rdoproject.org/r/c/openstack/tcib-distgit/+/47241) will be added later.

PR #16, opened 2023-02-28
Remove qdrouterd container

kajinamit:
The qdrouterd service was required by STF but will no longer be used, thus we can remove the container image.


rabi (approved):
lgtm

PR #17, opened 2023-02-28
[DNM] testing ci roles

raukadah:


rabi:
Does setting $ANSIBLE_ROLES_PATH to include that location not work?

rabi:
[ramishra@fedora tcib]$ANSIBLE_ROLES_PATH=./ci/roles ansible-playbook ci/playbooks/build_zed_containers.yaml  

This just works fine.. 

Also,  we need to remove the hack in the playbook.

https://github.com/rabi/tcib/blob/main/ci/playbooks/build_zed_containers.yaml#L9-L13

PR #21, opened 2023-03-01
Fix invalid tcib_actions

kajinamit:
tcib_actions should be a list.


rlandy (approved):
Tested this locally and it works.

PR #26, opened 2023-03-08
Add workflow to run tox jobs

kajinamit:
This change adds the jobs to run linter test and unit tests via tox when a change is pushed or a PR is created, so that we can verify the proposed change using these tests.


rabi (approved):
/lgtm

PR #31, opened 2023-03-29
Add exclude_containers option for osp18

rlandy:
This PR adds the osp18 exclude_containers
option and switches to use anchors to avoid
duplication.

### DIFF
--- a/ci/roles/build_containers/vars/main.yaml
+++ b/ci/roles/build_containers/vars/main.yaml
@@ -1,11 +1,11 @@
 ---
 exclude_containers:
   master:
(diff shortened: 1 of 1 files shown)


raukadah (approved):

kajinamit:
We removed neutron-mlnx-agent container by https://github.com/openstack-k8s-operators/tcib/pull/13 so we no longer need these excludes.

kajinamit:
I've created https://github.com/openstack-k8s-operators/tcib/pull/32 to remove these unnecessary exclusions.

PR #32, opened 2023-03-29
Remove redundant exclude_containers

kajinamit:
We have already removed neutorn-mlnx-agent so we don't need to exclude it.


rabi (approved):
/lgtm

PR #39, opened 2023-04-19
Fix missing ss command in ironic-pxe container

kajinamit:
The ironic-pxe container requires ss command to check status of UDP port used by dnsmasq.


rabi:
Probably good to add it in the base image like before.

rabi (approved):
/lgtm

PR #44, opened 2023-04-27
Set label tcib_build_tag to match tag

steveb:
This will allow the original build tag to be discovered by inspecting a stable tag such as current-podified. This may become useful later to replace stable tags with versioned tags for minor updates of running services.

### DIFF
--- a/tcib/builder/buildah.py
+++ b/tcib/builder/buildah.py
@@ -179,6 +179,7 @@ class BuildahBuilder(base.BaseBuilder):
         if self.debug:
             # TODO(bogdando): add --log-rusage for newer buildah
             bud_args.extend(['--loglevel=3'])
+        bud_args.extend(['--label', 'tcib_build_tag=%s' % self.tag])
         # TODO(aschultz): drop --format docker when oci format is properly
         # supported by the undercloud registry
         bud_args.extend(['--format', 'docker', '--tls-verify=False',
--- a/ci/roles/build_containers/defaults/main.yaml
+++ b/ci/roles/build_containers/defaults/main.yaml
@@ -35,7 +35,7 @@ release: "{{ ansible_distribution_major_version }}"
 # Skip or not the build of the images
 # skip_build
 
-# image tag
+# Image tag suffix. The image label tcib_build_tag will also be set with this value.
 image_tag: ''
 
 # Image Prefix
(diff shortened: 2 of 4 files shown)


tcib/builder/buildah.py
```diff
@@ -179,6 +179,7 @@ def build(self, container_name, container_build_path):
         if self.debug:
             # TODO(bogdando): add --log-rusage for newer buildah
             bud_args.extend(['--loglevel=3'])
+        bud_args.extend(['--label', 'tcib_build_tag=%s' % self.tag])
```
raukadah, line 182: Thank you for the pull request. Based on my understanding and 
 [code search ](https://github.com/search?q=repo%3Aopenstack-k8s-operators%2Ftcib%20label&type=code), It is already implemented in TCIB.

openstack tcib container image build ----label does the same.

Sorry I am not getting from this pr, can we add an example for the same.
TCIB also provides support for setting tags using --tag flag.

raukadah: Thank you steve for explaining it detail. Make sense now.

raukadah:
@steveb Can you add this parameter in the build_containers ansible role https://github.com/openstack-k8s-operators/tcib/blob/main/ci/roles/build_containers/defaults/main.yaml and https://github.com/openstack-k8s-operators/tcib/blob/main/ci/roles/build_containers/templates/build_containers.sh.j2 ?

It will be helpful for us to consume it in CI also.

steveb:
> @steveb Can you add this parameter in the build_containers ansible role https://github.com/openstack-k8s-operators/tcib/blob/main/ci/roles/build_containers/defaults/main.yaml and https://github.com/openstack-k8s-operators/tcib/blob/main/ci/roles/build_containers/templates/build_containers.sh.j2 ?
> 
> It will be helpful for us to consume it in CI also.

This change doesn't expose any new configuration interface, so I'll add some documentation for the `image_tag` parameter in `build_containers/defaults/main.yaml` to mention that this will also set label `tcib_build_tag`

steveb:
I have documented the `tcib_build_tag` label in `build_containers/defaults/main.yaml` and the `--tag` section of `openstack tcib container image build --help`.

raukadah (approved):

raukadah:
Thank you @steveb :-)

PR #46, opened 2023-05-11
Install python-tcib-containers package before it tries to

shreshtha90:
install scripts to make it available to be build by other build systems like ocds/brew where a job cannot pre-install the package


yazug (commented):
looks like that should do it, not sure how it was working before.

raukadah (approved):
Tested it here: https://review.rdoproject.org/r/c/testproject/+/47931 and https://review.rdoproject.org/zuul/build/9a76742e7f574dd29d8ad3b100d62fa6

raukadah:
> looks like that should do it, not sure how it was working before.
Due to https://github.com/openstack-k8s-operators/tcib/blob/main/ci/roles/build_containers/defaults/main.yaml#L51
```
# Container bind mount used when building the image
volume_mounts:
  - "/usr/share/tcib/container-images:/usr/share/tcib/container-images:z"
```
It was working.

PR #50, opened 2023-06-05
Require socat in mariadb image

dciabrin:
When galera is used with TLS, the SST method (rsync or mariabackup) requires encryption. This is currently achieved via a socat tunnel, so add socat as a dependency in the mariadb image.


rabi (approved):
Hmm.. It was initially there  and removed with https://github.com/openstack-k8s-operators/tcib/commit/ffe45c179875bd128627c5e90467de2b5863aa15 subsequently.

PR #52, opened 2023-06-06
Add tool to expose yum_update ansible tasks

tbreeds:
In some situtations it's desireable to update all RPMs in container that are from a specific repo.  This PR updates and exposes existing ansible tasks to do just that.

### DIFF
--- a/roles/modify_container_image/tasks/yum_update_buildah.yml
+++ b/roles/modify_container_image/tasks/yum_update_buildah.yml
@@ -110,7 +110,7 @@
       command: >
         buildah run
           --volume {{ yum_update.path }}:/tmp/yum_update.sh:z
-          --volume {{ yum_repos_dir_path }}:/etc/yum.repos.d
+          --volume {{ yum_repos_dir_path }}:/etc/yum.repos.d{{ ':z' if yum_repos_dir_path != "/etc/yum.repos.d"}}
           {% if pkg_mgr_vars_stat.stat.exists %}
           --volume /etc/{{ pkg_mgr_suffix }}/vars:/etc/{{ pkg_mgr_suffix }}/vars
           {% endif %}
--- a/roles/modify_container_image/tasks/yum_update_buildah.yml
+++ b/roles/modify_container_image/tasks/yum_update_buildah.yml
@@ -1,4 +1,7 @@
 ---
+# NOTE(tonyb):
+# This by default creates an: application/vnd.oci.image.config.v1+json image
+# not a application/vnd.oci.image.manifest.v1+json.
 - import_tasks: precheck.yml
   tags:
     - always
--- a/tcib/client/container_image.py
+++ b/tcib/client/container_image.py
@@ -826,3 +826,116 @@ class HotFix(command.Command):
                 playbook_dir=tmp,
                 verbosity=utils.playbook_verbosity(self=self),
             )
+
+
+class Update(command.Command):
+    """Update container images with tcib.
+
+    This differs from HotFix as we do not know ahead of time exactly which RPMs
+    will get updated, instead we rely on yum/dnf to make those decisions."""
+
+    log = logging.getLogger(__name__ + ".Update")
+    auth_required = False
+
+    def get_parser(self, prog_name):
+        parser = super(Update, self).get_parser(prog_name)
+        parser.add_argument(
+            "--image",
+            dest="images",
+            metavar="<images>",
+            default=[],
+            action="append",
+            required=True,
+            help=_(
+                "Fully qualified reference to the source image to be "
+                "modified. Can be specified multiple times (one per "
+                "image) (default: %(default)s)."
+            ),
+        )
+        parser.add_argument(
+            "--repos-path",
+            dest="repos_path",
+            metavar="<repos-path>",
+            default=os.path.join(os.getcwd(), "yum.repos.d"),
+            required=True,
+            help=_("Path containing the yum/dnf repo files "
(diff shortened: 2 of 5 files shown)


roles/modify_container_image/tasks/yum_update_buildah.yml
```diff
@@ -107,7 +110,7 @@
       command: >
         buildah run
           --volume {{ yum_update.path }}:/tmp/yum_update.sh:z
-          --volume {{ yum_repos_dir_path }}:/etc/yum.repos.d
+          --volume {{ yum_repos_dir_path }}:/etc/yum.repos.d:z
```
kajinamit, line 113: I vaguely remember we intentionally removed :z to avoid chcon over system /etc/yum.repos.d in case that directory is used.

kajinamit (changes requested):

roles/modify_container_image/tasks/yum_update_buildah.yml
tbreeds: Okay, I'll add a test so that if `yum_repos_dir_path` is `/etc/yum.repos.d` the "z" isn't added.

tcib/client/container_image.py
```diff
@@ -825,3 +826,111 @@ def take_action(self, parsed_args):
...
+        parser.add_argument(
+            "--repos-path",
+            dest="repos_path",
+            metavar="<repos-path>",
+            required=True,
+            help=_("Path containing the yum/dnf repo files "
+                   "(default: %(default)s)."),
+        )
+        parser.add_argument(
+            "--update_repo",
```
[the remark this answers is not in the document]
tbreeds: Yup for sure I can fix that ... the '_' is a mistake.

As for the name. I was following the naming in [roles/modify_container_image/tasks/yum_update_buildah.yml L126](https://github.com/openstack-k8s-operators/tcib/blob/main/roles/modify_container_image/tasks/yum_update_buildah.yml#L126) etc.

Personally I'd find it confusing for gthe names not to match.  I can change it in in `yum_update_buildah` but that might break existing users .... Not that I think there are any.

Thoughts?

tcib/client/container_image.py
```diff
@@ -825,3 +826,111 @@ def take_action(self, parsed_args):
...
+                "image) (default: %(default)s)."
+            ),
+        )
+        parser.add_argument(
+            "--repos-path",
+            dest="repos_path",
+            metavar="<repos-path>",
+            required=True,
+            help=_("Path containing the yum/dnf repo files "
+                   "(default: %(default)s)."),
```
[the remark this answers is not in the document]
tbreeds: Fixed:  I'll add:

`default=os.path.join(os.getcwd(), "yum.repos.d"),`  in my next push

roles/modify_container_image/tasks/yum_update_buildah.yml
tbreeds: For the next itteration of this PR I'm going to leave the ':z' in place as I can't confirm that recollection:

```
[tony@thor tcib]$ git -C /home/tony/src/github.com/openstack-k8s-operators/tcib log origin/main -- roles/modify_container_image/tasks/yum_update_buildah.yml
commit 11a1ff48bfc0fdcc8a8a984d42507f5215ccd5a5
Author: Chandan Kumar <raukadah@gmail.com>
Date:   Thu Feb 23 11:12:09 2023 +0530

    Import modify image role in tcib
    
    It will be used with openstack tcib container image hotfix
    command.
    
    Signed-off-by: Chandan Kumar <raukadah@gmail.com>
```
and the "source" project has a change very similar to mine:
```
[tony@thor tcib]$ git -C /tmp/ansible-role-tripleo-modify-image log -1 origin/master -- tasks/yum_update_buildah.yml
commit b6eedb6fb194ec32b92866257131714e02bae741 (HEAD -> master, origin/master, origin/HEAD)
Author: Cédric Jeanneret <cjeanner@redhat.com>
Date:   Wed Jan 18 16:26:01 2023 +0100

    Ensure update_yum.sh has correct SELinux labels
    
    On recent, enforcing systems, the update_yum.sh can't be executed,
    because SELinux prevents container_t to open user_tmp_t:
    
    type=AVC msg=audit(1674049913.380:22858): avc:  denied  { open } for
        pid=70472 comm="bash" path="/tmp/yum_update.sh" dev="vda4" ino=218200014
        scontext=system_u:system_r:container_t:s0:c65,c705
        tcontext=unconfined_u:object_r:user_tmp_t:s0
        tclass=file permissive=0
    
    This patch ensures it gets properly relabelled when bind-mounted during
    the image build. Using the "z" will also ensure it's still usable even
    when running multiple builds at the same time.
    
    Change-Id: I4085865965f48c9fa6a88cde7010a51cd8c653d8
```

kajinamit: No this change fixes the selinux error about the yum_update.sh script. I double-checked the current master of ansible-role-tripleo-modify-image it does not have z to mount repository directory.

https://github.com/openstack/ansible-role-tripleo-modify-image/blob/master/tasks/yum_update_buildah.yml#L110

I'd suggest we copy `/etc/yum.repos.d` to a different path (eg `/tmp/yum.repos.d`) and mount that directory if we really need :z flag added.

kajinamit (approved):
not too sure if we can remove z really in case `/etc/yum.repos.d` is used and I still feel we need to create a copy directory to make the command work with that default path, but if this already meets the requirement then that'd be overkilling.

PR #54, opened 2023-07-10
Install glanceclient

konan-abhi:
This ensures we install glanceclient library so that we can use additional commands for glance service which are not supported/available in openstackclient.


kajinamit:
Currently users are supposed to launch a container using the OpenStackClient CR and execute clients in that pod. Looking at the current implementation we only expose clouds.yaml and do not provide any RC file or env vars.
If we expect users need to use the legacy client very frequently then we should probably consider adding mechanism to expose auth vars.

fmount (approved):

kajinamit:
Cinder folks might be interested in adding their legacy client(cinderclient) as well

PR #55, opened 2023-07-14
Fix default container namespace

karelyatin:
Containers are built and pushed to namespace
podified-main-centos9, So let's use that instead.


kajinamit:
IMO `master` should be correct here, because these branch names should be consistent with upstream OpenStack projects.

karelyatin:
No longer required, container images are now using master namespace.

PR #56, opened 2023-07-19
Source container-images from the build host, not an rpm

steveb:
It is currently not practical to do local tcib development and build images with those changes, because the container-images directory is sourced from an rpm install inside the container, instead of the container-images from the build host.

With this change, --config-path and TCIB_CONFIG_PATH will now be honoured, so local builds will work as expected.

This may also fix weird pipeline issues where published container images don't get changes until rebuilt with the published python-tcib-containers rpm.


steveb:
I would add that the python-tcib-containers rpm is still useful, for setting up build hosts with appropriate tcib

rabi (commented):
I'm fine with this. Probably good to check with RelDel before merging.

raukadah:
@yazug please have a look at this change, It might break downstream container builds.

yazug:
This will break downstream builds.  Is there a way to guard from local or from RPM that would allow the right solution to be picked up by systems/usage?

raukadah:
> This will break downstream builds. Is there a way to guard from local or from RPM that would allow the right solution to be picked up by systems/usage?

Thank you @yazug for looking at the pr. Based on the discussion with @steveb If we install python-tcib-containers on the host machine, then it will install all the container-images content in /usr/share directory. 

On Running tcib with this pr, we will be mounting /usr/share volume directory (which containers python-tcib-containers content) while building the containers. This way we will not miss any stuff from python-tcib-containers rpm. And it will not break downstream build.

yazug:
So issue is that when building container images downstream with brew (koji) container-build ... there is no host machine and we build from Dockerfile .  If the tcib rpm is not installed in the container there will be no files to copy into place.

steveb:
I've reworked this so there is no change in behaviour with the default arguments, rendered Dockerfiles will still have the required package.

A working development mode has been documented, which is basically clearing the `tcib_package` variable

yazug:
after the rework it looks like this should work for downstream.  Thanks much!

rabi (approved):
/lgtm

PR #61, opened 2023-07-27
openstackclient improvements

steveb:
- Better set up cloud-admin user in openstackclient
- Add iputils to openstackclient
- Add missing baremetal command completion


hjensas (approved):
LGTM.

container-images/tcib/base/openstackclient/openstackclient.yaml
abays, line 1: nit: no newline at end of file.  Otherwise looks good.

container-images/tcib/base/openstackclient/openstackclient.yaml
```diff
@@ -1,7 +1,10 @@
 tcib_actions:
-- run: bash /usr/local/bin/uid_gid_manage {{ tcib_user }}
+- run: rm -rf /home/cloud-admin && bash /usr/local/bin/uid_gid_manage {{ tcib_user }}
```
kajinamit, line 2: We earlier discussed how we get rid of the usage of cloud-admin user with fixed uid/gid, and we probably have to reconsider this when we use arbitrary uid/gid assigned  by k8s, but I'm ok we merge this in the current implementation for now.

PR #66, opened 2023-08-14
creating tempest-extras container

afazekas:
The container will include the tempest(or other) plugins
with has a public source, but does not shipped as rpm,
so from git repository.

### DIFF
--- a/container-images/tcib/base/os/tempest/tempest-extras/tempest-extras.yaml
+++ b/container-images/tcib/base/os/tempest/tempest-extras/tempest-extras.yaml
@@ -1,0 +1,10 @@
+tcib_actions:
+- run: dnf -y install {{ tcib_packages['common'] | join(' ') }} && dnf clean all && rm -rf /var/cache/dnf
+- run: cd /usr/local; git clone https://opendev.org/openstack/whitebox-tempest-plugin
+- run: cd /usr/local/whitebox-tempest-plugin; pip install -e .
+
+tcib_packages:
+  common:
+  - python3-pip
+
+tcib_user: tempest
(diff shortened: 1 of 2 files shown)


container-images/tcib/base/os/tempest/tempest-extras/tempest-extras.yaml
```diff
@@ -0,0 +1,10 @@
+tcib_actions:
+- run: dnf -y install {{ tcib_packages['common'] | join(' ') }} && dnf clean all && rm -rf /var/cache/dnf
+- run: cd /usr/local; git clone https://opendev.org/openstack/whitebox-tempest-plugin
```
raukadah, line 3: whitebox-tempest-plugin is already packaged in rdo: https://github.com/rdo-packages/whitebox-tempest-plugin-distgit Do we still want to use whitebox-tempest-plugin here?

jamepark4: Yes compute is definitely interested in finding a way to incorporate the whitebox plugin into a tempest container.  Either with this approach or if we can find a way in the build process to add a whitebox (or some other plugin) layer to the container that we can reference when testing.

raukadah: Thank you @jamepark4 , Let's give it a go.

raukadah (approved):

afazekas:
ATM I see python3-whitebox-neutron-tests-tempest.noarch in the repos (https://github.com/rdo-packages/whitebox-neutron-tempest-plugin-distgit), but not $subject package.

raukadah:
@afazekas Recently we have created rdo branches for whitebox-neutron-tempest-plugin in order to test with wallaby tripleo jobs https://review.rdoproject.org/r/c/config/+/49319. It will make the package available for most of the supported release . If we want to test it via package do let me know, I can make it available.

PR #67, opened 2023-08-23
Remove oslo.vmware from nova/glance

kajinamit:
The vmwareapi virt driver was deprecated a while ago[1] because of lack of CI and maintainers. Because we haven't seen any users attempting to use the driver even in TripleO project, let's remove the package to reduce image size.

[1] https://review.opendev.org/c/openstack/nova/+/863911

### DIFF
--- a/container-images/tcib/base/os/glance-api/glance-api.yaml
+++ b/container-images/tcib/base/os/glance-api/glance-api.yaml
@@ -9,7 +9,6 @@ tcib_packages:
   - httpd
   - mod_ssl
   - openstack-glance
-  - python3-oslo-vmware
   - python3-rados
   - python3-rbd
   - qemu-img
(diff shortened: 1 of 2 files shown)


konan-abhi:
No concerns from glance!

gibizer (commented):
upstream nova declares that oslo.vmware is an optional dependency. https://github.com/openstack/nova/blob/17b46bae84bef38bc27976219a244010f1e510e3/setup.cfg#L36-L37  So nova is expected to work without that package if the vmware virt driver is not used. In openstack-k8s-operators at the moment we only support the libvirt virt driver (and support comming for ironic and the fake virt driver) but I don't think it is on our roadmap to support the vmware virt driver. So I think it is OK to remove that package

gibizer (approved):

PR #68, opened 2023-08-23
Remove unused image and pmem deps

SeanMooney:
nova upstream support persitent memory but
intel discontinued the product line and downstream
support has been removed form our product.
as a result the daxio and ndctl binaries are nolonger used.

This change removes them and the unused nova-compute-ironic image.
The normal nova-compute image will be used for ironic, libvirt and
the fake driver so the nova-compute-ironic image is not required.

Closes: [OSPRH-136](https://issues.redhat.com//browse/OSPRH-136)


container-images/tcib/base/os/nova-base/nova-compute/nova-compute.yaml
```diff
@@ -27,6 +26,4 @@ tcib_packages:
   - swtpm-tools
   - targetcli
   - xfsprogs
-  x86_64:
```
kajinamit, line 30: You have to remove the reference in line 2 as well.

kajinamit (changes requested):
Looks like the failure in container build job is real and needs a small update.

Also, do you mind fixing a typoe in the title ? (s/Remote/Remove/)

container-images/tcib/base/os/nova-base/nova-compute/nova-compute.yaml
SeanMooney: ah yes i did not se that conditional install, will do.

kajinamit (approved):

PR #71, opened 2023-08-28
Gnocchi: Install apache in api container only

kajinamit:
apache is required only by the api container and is not used by the other containers such as metricd.


mrunge:
This change makes sense to me.

raukadah (approved):

PR #72, opened 2023-08-28
Ceilometer: Remove unused oslo.db

kajinamit:
Ceilometer no longer uses oslo.db, since it offloaded its metric storage and event storage.

Also explicit install of tooz library is not required because the package is part of dependencies and should be installed automatically.

### DIFF
--- a/container-images/tcib/base/os/ceilometer-base/ceilometer-base.yaml
+++ b/container-images/tcib/base/os/ceilometer-base/ceilometer-base.yaml
@@ -4,5 +4,3 @@ tcib_actions:
 tcib_packages:
   common:
   - openstack-ceilometer-common
-  - python3-oslo-db
-  - python3-tooz


mrunge:
tooz is still used in ceilometer https://github.com/openstack/ceilometer/blob/master/ceilometer/polling/manager.py#L34

kajinamit:
> tooz is still used in ceilometer https://github.com/openstack/ceilometer/blob/master/ceilometer/polling/manager.py#L34

tooz is still used but the package should be installed according to the dependency automatically so we don't need to explicitly define it.
 https://github.com/rdo-packages/ceilometer-distgit/blob/antelope-rdo/openstack-ceilometer.spec#L116

kajinamit:
recheck

404 while downloading repository metadata

raukadah (approved):

PR #76, opened 2023-09-07
Add hotfix role

son-vyas:
Addition of this role automates the process of applying hotfixes to container images.

### DIFF
--- a/roles/hotfix/defaults/main.yml
+++ b/roles/hotfix/defaults/main.yml
@@ -1,0 +1,6 @@
+---
+# Name of the container image to be hotfixed
+container_image_name: ""
+
+# Path to the hotfix script
+hotfix_script_path: ""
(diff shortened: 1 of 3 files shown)


roles/hotfix/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
```
raukadah, line 1: I think this role will be added in the ci-framework not in tcib repo.

roles/hotfix/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
...
+# Name of the container image to be hotfixed
+container_image_name: ""
```
raukadah, line 3: We have copied the hotfix code from tripleo-common. I have tried this interface. I think we can follow the tripleo docs: https://docs.openstack.org/project-deploy-guide/tripleo-docs/latest/deployment/container_image_prepare.html#building-hotfixed-containers and check what interface of hotfix works and add a example in the tcib. It will help to create the new role.

son-vyas: Should we close this pr and have new one for adding example in the tcib?

son-vyas: i am closing this pr since we no longer require this

PR #80, opened 2023-09-14
Fix run_tempest.sh

arxcruz:
The concurrency variable wasn't being set properly, as well as the check if the concurrency was set, making tempest jobs start to fail. This patch fixes it, and also set the shell to show the output of all commands, making it easier in the future to debug any failure.


viroel:
/lgtm

thanks for the fix

viroel:
/approve

fixes the issue with latest promoted tempest container.

PR #94, opened 2023-10-24
Fix dependency handling logic

olliewalsh:
The note[1] indicates that dependencies will be built but this
only appears to work for the first ancestor of an image.
Fix this by walking up the directory tree for a given image to
find all ancestors to include in the build.

Also record the level of each image and process in a breadth first order,
removing the hardcoded ordering for the 'base' and 'os' images.

[1] https://github.com/openstack-k8s-operators/tcib/blob/41a73ab686826c4fc2095cbe3f76b0dc2d61147f/tcib/client/container_image.py#L537-L541

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,14 +23,13 @@ $ sudo dnf -y install ansible-core buildah
 
 3. Define subset of images to build
 
-By default all images will be built as specified
+By default all images and their dependencies will be built as specified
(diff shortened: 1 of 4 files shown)


stuggi (commented):
looks sane to me, but would be good if someone who is more familiar with the build process to check it.

raukadah (approved):
Looks good, thank you @olliewalsh for improving it. :-)

PR #95, opened 2023-10-25
Add support for tempestconf parameters

lpiwowar:
This patch introduces tempestconf parameters for the tempest container. Currently, it is not possible to specify with what parameters should be tempest and discover-tempest-config executed.

This patch introduces a set of environment variables prefixed with TEMPESTCONF_  that can be used to tell the container which arguments should be used when executing discover-tempest-config.

### DIFF
--- a/container-images/tcib/base/os/tempest/run_tempest.sh
+++ b/container-images/tcib/base/os/tempest/run_tempest.sh
@@ -1,11 +1,50 @@
 #!/bin/sh
-
 set -x
 
 HOMEDIR=/var/lib/tempest
 TEMPEST_DIR=$HOMEDIR/openshift
 PROFILE_ARG=""
 CONCURRENCY="${CONCURRENCY:-}"
+
+TEMPESTCONF_ARGS=""
+
+[[ ${TEMPESTCONF_CREATE:=true} == true ]] && TEMPESTCONF_ARGS+="--create "
+[[ ${TEMPESTCONF_INSECURE} == true ]] && TEMPESTCONF_ARGS+="--insecure "
+[[ ${TEMPESTCONF_COLLECT_TIMING} == true ]] && TEMPESTCONF_ARGS+="--collect-timing "
+[[ ${TEMPESTCONF_NO_DEFAULT_DEPLOYER} == true ]] && TEMPESTCONF_ARGS+="--no-default-deployer "
+[[ ${TEMPESTCONF_DEBUG:=true} == true ]] && TEMPESTCONF_ARGS+="--debug "
+[[ ${TEMPESTCONF_VERBOSE} == true ]] && TEMPESTCONF_ARGS+="--verbose "
+[[ ${TEMPESTCONF_NO_RNG} == true ]] && TEMPESTCONF_ARGS+="--no-rng "
+[[ ${TEMPESTCONF_NON_ADMIN} == true ]] && TEMPESTCONF_ARGS+="--non-admin "
+[[ ${TEMPESTCONF_RETRY_IMAGE} == true ]] && TEMPESTCONF_ARGS+="--retry-image "
+[[ ${TEMPESTCONF_CONVERT_TO_RAW} == true ]] && TEMPESTCONF_ARGS+="--convert-to-raw "
+
+[[ ! -z ${TEMPESTCONF_TIMEOUT} ]] && TEMPESTCONF_ARGS+="--timeout ${TEMPESTCONF_TIMEOUT} "
+[[ ! -z ${TEMPESTCONF_OUT} ]] && TEMPESTCONF_ARGS+="--out ${TEMPESTCONF_OUT} "
+[[ ! -z ${TEMPESTCONF_DEPLOYER_INPUT} ]] && TEMPESTCONF_ARGS+="--deployer-input ${TEMPESTCONF_DEPLOYER_INPUT} "
+[[ ! -z ${TEMPESTCONF_TEST_ACCOUNTS} ]] && TEMPESTCONF_ARGS+="--test-accounts ${TEMPESTCONF_TEST_ACCOUNTS} "
+[[ ! -z ${TEMPESTCONF_CREATE_ACCOUNTS_FILE} ]] && TEMPESTCONF_ARGS+="--create-accounts-file ${TEMPESTCONF_CREATE_ACCOUNTS_FILE} "
+[[ ! -z ${TEMPESTCONF_PROFILE} ]] && TEMPESTCONF_ARGS+="--profile ${TEMPESTCONF_PROFILE} "
+[[ ! -z ${TEMPESTCONF_GENERATE_PROFILE} ]] && TEMPESTCONF_ARGS+="--generate-profile ${TEMPESTCONF_GENERATE_PROFILE} "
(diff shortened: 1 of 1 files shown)


lpiwowar:
Related patch for test-operator: https://github.com/openstack-k8s-operators/test-operator/pull/10

arxcruz (commented):
I am not sure, but I believe this will break the actual tempest, did you tested locally? I am not seeing some options that was hardcoded there...

lpiwowar:
Thank you for looking at the patch Arx. Yes, I ran the changes locally and they seemed to be working. The options that were hardcoded are still enabled by default ([line 10](https://github.com/openstack-k8s-operators/tcib/pull/95/commits/74139bd57b6dc5586666696fc59a7fc3804e453e#diff-191bc7762340f2539b035115667e34a49916229d47d776aca4a2c457f694e35cR10), [line 14](https://github.com/openstack-k8s-operators/tcib/pull/95/commits/74139bd57b6dc5586666696fc59a7fc3804e453e#diff-191bc7762340f2539b035115667e34a49916229d47d776aca4a2c457f694e35cR14) and [line 46](https://github.com/openstack-k8s-operators/tcib/pull/95/commits/74139bd57b6dc5586666696fc59a7fc3804e453e#diff-191bc7762340f2539b035115667e34a49916229d47d776aca4a2c457f694e35cR46)). The only option that was removed is the --os-cloud but I think it should not be a problem as the OS_CLOUD environment variable (which should be detected by python-tempestconf) is defined in the script.

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -18,15 +56,10 @@ if [ ! -z ${USE_EXTERNAL_FILES} ]; then
     cp ${TEMPEST_PATH}clouds.yaml $HOME/.config/openstack/clouds.yaml
 fi
 
-if [ -f ${TEMPEST_PATH}profile.yaml ]; then
```
kopecmartin, line 21: if we delete this block, we'll break the ci-framework usage of this image ... they generate profile.yaml on the fly - https://github.com/openstack-k8s-operators/ci-framework/blob/56d9e69a2ce00d96dcf006f86b088ee944c1a39d/ci_framework/roles/tempest/tasks/configure-tempest.yml#L34 

Then the file would be discovered by this if-statement and passed to tempestconf, if we remove it, the generated profile.yaml file will not make it to tempestconf.

I see couple of options:
1. set TEMPESTCONF_PROFILE somewhere in ci-framework/roles/tempest
2. extend the logic of this if-statement so that if TEMPESTCONF_PROFILE is not set,  ${TEMPEST_PATH}profile.yaml will be passed to --profile if the file exists.

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -18,15 +56,10 @@ if [ ! -z ${USE_EXTERNAL_FILES} ]; then
...
-    PROFILE_ARG="--profile ${TEMPEST_PATH}profile.yaml"
-fi
-
 tempest init openshift
 
 pushd $TEMPEST_DIR
-
-discover-tempest-config --os-cloud $OS_CLOUD --debug --create identity.v3_endpoint_type public ${PROFILE_ARG}
```
kopecmartin, line 29: --os-cloud $OS_CLOUD
^^ if we wanna maintain the same logic, if we remove it from here, we need to set this here: we need to set this in https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/tempest/vars/main.yml so that it remains a default value (but this way it'll be customizable if needed)

--debug
^^ we need to set this in https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/tempest/vars/main.yml

--create
^^ we need to set this in https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/tempest/vars/main.yml

identity.v3_endpoint_type public
^^ this is ok to remove as it's set ci-framework by default via profile.yaml (that's the only tempestconf interface that ci-framework uses to interact with temepstconf) -> https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/tempest/vars/main.yml

kopecmartin (changes requested):

container-images/tcib/base/os/tempest/run_tempest.sh
lpiwowar: That's a good point. 

What do you think about sticking only with TEMPESTCONF_PROFILE and removing the option of passing the discover-tempest-config parameters by mounting the profile.yaml to the image? I would stick with only one interface. Keeping it simple. We are still in the early stages so I think we can afford it if we modify ci-framework correctly (?). Just an idea.

container-images/tcib/base/os/tempest/run_tempest.sh
kopecmartin: ah, ok, I see, I missed that you set true by default on lines 10 and 14 so `--debug` and `--create` are covered, the logic is not changed. 

I'd personally avoid setting default values in this script but that can be changed later (e.g. after this is merged, we can propose changes to ci-framework where we can set --debug and --create by default and after that is merged we could remove the default values here from lines 10 and 14). The same applies on the `identity.v3_endpoint_type public` override.

I agree with your comments about OS_CLOUD.

container-images/tcib/base/os/tempest/run_tempest.sh
kopecmartin: I'm afraid I don't follow. As it is right now, there are 2 options how to pass arguments to tempestconf.
1. generate a custom profile.yaml that is mounted to the container - that's how ci-framework sets options for discover-tempest-config right now - https://github.com/openstack-k8s-operators/ci-framework/blob/d776dc8babe7117c2d29ad722b70117da1bde09a/ci_framework/roles/tempest/tasks/configure-tempest.yml#L34
2. set desired `TEMPESTCONF_*` variables

If you remove this if-statement, the profile.yaml file that is generated by ci-framework will be ignored unless we set `TEMPESTCONF_PROFILE` in ci-framework.

kopecmartin:
the changes look ok to me and are backwards compatible now

kopecmartin:
hmm, why it still shows I requested changes? I'd say that should go away after a new commit was proposed (well, force pushed in this case). Hmm, what if I click on Approve changes? Will that make the "Changes requested" message go away or will that merge this?

kopecmartin (approved):

lpiwowar:
recheck

kopecmartin:
recheck

raukadah:
We need to wait for https://github.com/openstack-k8s-operators/ci-framework/pull/785 and https://github.com/openstack-k8s-operators/tcib/pull/97 to land for passing zuul jobs.

kopecmartin:
recheck

raukadah:
recheck

raukadah:
recheck

kopecmartin:
the CI estimates more then 9 hours till the jobs finish and (maybe) merge this - https://review.rdoproject.org/zuul/status - 9 hours :o so much for shortening the feedback loops :D

kopecmartin:
recheck

raukadah (approved):

PR #96, opened 2023-10-26
Add tempest parameters for the tempest container

lpiwowar:
This patch introduces tempest parameters for the tempest container. Currently, it is not possible to specify with what parameters should be tempest executed.

This patch introduces a set of environment variables prefixed with TEMPEST_ that can be used to tell the container which arguments should be used when executing tempest.

### DIFF
--- a/container-images/tcib/base/os/tempest/run_tempest.sh
+++ b/container-images/tcib/base/os/tempest/run_tempest.sh
@@ -3,10 +3,9 @@ set -x
 
 HOMEDIR=/var/lib/tempest
 TEMPEST_DIR=$HOMEDIR/openshift
-PROFILE_ARG=""
 CONCURRENCY="${CONCURRENCY:-}"
-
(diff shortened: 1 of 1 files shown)


container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -4,8 +4,17 @@ set -x
 
 HOMEDIR=/var/lib/tempest
 TEMPEST_DIR=$HOMEDIR/openshift
-PROFILE_ARG=""
-CONCURRENCY="${CONCURRENCY:-}"
```
afazekas, line 8: You may want to deprecate the old org, but keep it functional for a while.

lpiwowar: Ok, understood. Fixed:)

raukadah:
recheck

raukadah (approved):

PR #103, opened 2023-11-21
Support external tempest plugins

lpiwowar:
This patch adds support for the tempest container to be able to install external plugins that are not part of the .rpms the container is built with.

When a user defines the TEMPEST_EXTERNAL_PLUGIN_GIT_URLS then plugin(s) defined by this variable will be downloaded and executed. The user can also define TEMPEST_EXTERNAL_PLUGIN_CHANGE_URLS and TEMPEST_EXTERNAL_PLUGIN_REFSPEC variable to download a change(s) for a given plugin. This can be used for testing tests under developoment.

TEMPEST_EXTERNAL_PLUGIN_GIT_URLS
    - URL of a git repository that holds the plugin

TEMPEST_EXTERNAL_PLUGIN_CHANGE_URLS
    - URL from which a change should be downloaded for a given plugin

TEMPEST_EXTERNAL_PLUGIN_REFSPEC
    - A change that should be downloaded for a given plugin


kopecmartin:
for anyone looking at this PR, the documentation on how this feature would be used is added by this PR (and still being discussed):
https://github.com/openstack-k8s-operators/test-operator/pull/12

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -2,6 +2,7 @@
 set -x
 
 HOMEDIR=/var/lib/tempest
+TEMPEST_PATH=$HOMEDIR/
```
kopecmartin, line 5: duplicate of line 7

kopecmartin:
is this rebased on the latest changes?

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -2,8 +2,8 @@
 set -x
 
 HOMEDIR=/var/lib/tempest
-TEMPEST_DIR=$HOMEDIR/openshift
 TEMPEST_PATH=$HOMEDIR/
+TEMPEST_DIR=$HOMEDIR/openshift
```
kopecmartin, line 6: unrelated change (moved from line 5 to 6)

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -34,7 +34,22 @@ TEMPEST_ARGS=""
 [[ ! -z ${TEMPESTCONF_FLAVOR_MIN_DISK} ]] && TEMPESTCONF_ARGS+="--flavor-min-disk ${TEMPESTCONF_FLAVOR_MIN_DISK} "
 [[ ! -z ${TEMPESTCONF_NETWORK_ID} ]] && TEMPESTCONF_ARGS+="--network-id ${TEMPESTCONF_NETWORK_ID} "
 
+TEMPESTCONF_OVERRIDES="$(echo ${TEMPESTCONF_OVERRIDES} | tr '\n' ' ') identity.v3_endpoint_type public"
```
kopecmartin, line 37: is this change required? The line's been moved from line 61 to 37

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -34,7 +34,22 @@ TEMPEST_ARGS=""
...
+
 # Tempest arguments
+TEMPEST_EXTERNAL_PLUGIN_GIT_URL="${TEMPEST_EXTERNAL_PLUGIN_GIT_URL:-}"
+TEMPEST_EXTERNAL_PLUGIN_CHANGE_URL="${TEMPEST_EXTERNAL_PLUGIN_CHANGE_URL:-}"
+TEMPEST_EXTERNAL_PLUGIN_REFSPEC="${TEMPEST_EXTERNAL_PLUGIN_REFSPEC:-}"
+TEMPEST_EXTERNAL_PLUGIN_DIR=/var/lib/tempest/external-plugins
+
+# Convert comma separated lists to arrays
+OLD_IFS=$IFS
```
kopecmartin, line 46: just out of curiosity, what does this stand for?

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -58,13 +73,82 @@ if [[ ! -z ${TEMPESTCONF_REMOVE} ]]; then
...
+
+        if [[ ! -z $change_url ]] && [[ ! -z $refspec ]] || \
+           [[ $change_url != "-" ]] && [[ $refspec != "-" ]]; then
+            pushd $plugin_name
+            git fetch $change_url $refspec
+            git checkout FETCH_HEAD
+            popd
+        fi
+
+        pip install ./$plugin_name
```
kopecmartin, line 104: pip/pip3? does it make a difference? I guess pip points to pip3 so it doesn't matter

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -58,13 +73,82 @@ if [[ ! -z ${TEMPESTCONF_REMOVE} ]]; then
...
+    done
+
+    pushd $HOMEDIR
+    tempest init openshift
+    pushd $TEMPEST_DIR
+
+    discover-tempest-config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES}
+    tempest run ${TEMPEST_ARGS}
```
kopecmartin, line 112: hmm, what tempest is used here? the one installed from rpm? what if we wanna install tempest from source (e.g. to test a new patch etc). I would assume that if we pass tempest source as another plugin, tempest installed in the venv would override the globally installed one, so it should work. Any chance you tested that so that we're sure?

container-images/tcib/base/os/tempest/run_tempest.sh
lpiwowar: Not really. But I'm trying to follow the pattern of having all the one-liners that depend on external variables at the beginning of the script. Should I move it back?

container-images/tcib/base/os/tempest/run_tempest.sh
lpiwowar: IFS is [Internal Field Separator](https://www.baeldung.com/linux/ifs-shell-variable)

container-images/tcib/base/os/tempest/run_tempest.sh
kopecmartin: nah, it's fine

container-images/tcib/base/os/tempest/run_tempest.sh
kopecmartin: right! I totally forgot some bash details :/ :D

container-images/tcib/base/os/tempest/run_tempest.sh
lpiwowar: I think that using pip should be ok here. We are inside a virtual environment.

container-images/tcib/base/os/tempest/run_tempest.sh
lpiwowar: hmm, what tempest is used here? the one installed from rpm?

- Either tempest that is installed as a dependency for a plugin
- or a tempest that is installed separately because a user specified it using the newly introduced variables in this patch.

Any chance you tested that so that we're sure?
- I hadn't tested it before I read this comment. I think you made a good point. But I've just tested it and the feature seems to be working even when a user specifies tempest via the newly introduced variables.

kopecmartin: thanks

kopecmartin (approved):
lgtm

kopecmartin:
recheck

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -58,13 +73,82 @@ if [[ ! -z ${TEMPESTCONF_REMOVE} ]]; then
...
+    for plugin_index in "${!TEMPEST_EXTERNAL_PLUGIN_GIT_URL[@]}"; do
+        git_url=${TEMPEST_EXTERNAL_PLUGIN_GIT_URL[plugin_index]}
+        change_url=${TEMPEST_EXTERNAL_PLUGIN_CHANGE_URL[plugin_index]}
+        refspec=${TEMPEST_EXTERNAL_PLUGIN_REFSPEC[plugin_index]}
+        plugin_name=$(basename -s .git $git_url)
+
+        git clone $git_url
+
+        if [[ ! -z $change_url ]] && [[ ! -z $refspec ]] || \
+           [[ $change_url != "-" ]] && [[ $refspec != "-" ]]; then
```
marios, line 97: maybe a comment would help but i'm trying to imagine what $change_url looks like here and why we need to guard for != "-" (and failing to imagine it :))

marios (commented):
is this tested somewhere (eg with testproject) with passing the new vars in?

feels like run_tempest.sh needs a README or at least some comments at the top (like the information you have in your commit message) to document what these vars are/do

container-images/tcib/base/os/tempest/run_tempest.sh
lpiwowar: I explained the usage of the variables used here in the comment at the beginning of the script. I can imagine that the "-" is a little bit confusing. I think that it would be used only in edge cases when someone would want to download multiple changes.

I think that the conversation here https://github.com/openstack-k8s-operators/test-operator/pull/12 can be helpful as well.

son-vyas:
any issue in testing with testproject ? like without testing this its hard to approve

kopecmartin:
> any issue in testing with testproject ? like without testing this its hard to approve

tested locally, no testproject, these changes are required in test-opeartor - https://github.com/openstack-k8s-operators/test-operator/pull/12 - which isn't used anywhere yet as we're writing an ansible role to run it in CI, still in progress. In this review it's enough if we check only that the image works with current CI jobs. I guess some have been executed when this was proposed. Based on my code review, the changes in the code don't affect the CI jobs as well as don't affect the current way how we use this image.

viroel:
I don't really remember where we replace the tempest image with the one built in content-provider. So yesterday I created a testproject to force tempest image to use the one built by the job. Job continues to pass.
https://review.rdoproject.org/r/c/testproject/+/48781/50#message-e9e7035f564f851a1df4e1cb7df9565a6f205b0a

Trying to run now again, without tempest_image var to see the output from podman images.

container-images/tcib/base/os/tempest/run_tempest.sh
viroel: I got it. Not sure if there is a better way to implement it, but I see that you need "-" because your loop gets values based on indexes.
A good documentation was added to the script itself. Is there another place that we can add this? So everybody would know how to use this?

viroel:
/approve

based on https://review.rdoproject.org/zuul/build/d2b3ab07e8f3416087de6da7f2cc9194 too. Since doesn't break edpm-multinode jobs.

PR #112, opened 2023-12-12
Remove temporary skip of tempest-all package

kopecmartin:
openstack-tempest-all package was temporarily replaced with only
openstack-tempest one because whitebox-neutron-tempest plugin
wasn't available on rhel.
These patches make sure that the plugin is included in the
openstack-tempest-all package only in upstream:
https://review.rdoproject.org/r/q/I93f6cee77927dfadd15c8093b49b1ed1f3e04cdc
Thus it's safe to get the things back as they were.

### DIFF
--- a/container-images/tcib/base/os/tempest/tempest-all/tempest-all.yaml
+++ b/container-images/tcib/base/os/tempest/tempest-all/tempest-all.yaml
@@ -14,12 +14,10 @@ tcib_actions:
 
 tcib_entrypoint: /var/lib/tempest/run_tempest.sh
 
-# TODO(arxcruz): ADd openstack-tempest-all again once tempest-whitebox
-# package is available on rhel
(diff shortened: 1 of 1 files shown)


viroel:
/approve

thanks, this is a must, we need plugins installed to continue with other tasks.

gouthampacha:
+1

We'll need this to run tests from tempest plugins in the jobs... Thanks @kopecmartin

arxcruz (approved):
/approve but if it start to fails, I will revert again :)

viroel:
> /approve but if it start to fails, I will revert again :)

It is worst to build image called `tempest-all` without `openstack-tempest-all` package, it causes confusion for those that are using it - it is better to stop building it or letting it failure until the correct fix is landed.

kopecmartin:
recheck

PR #113, opened 2023-12-13
[tempest-extras]Add barbican tempest plugin

gibizer:
The whitebox tempest plugin actually depends on the barbican tempest plugin otherwise the vTPM tests are failing with
```
    cls.os_primary.secrets_client = service_clients.secret_v1.SecretClient(
AttributeError: 'ServiceClients' object has no attribute 'secret_v1'
```
So this patch adding it.

Implements: https://issues.redhat.com/browse/OSPRH-2451

PR #115, opened 2023-12-15
[tempest-all]Add whitebox-tempest-plugin

gibizer:
Implements: https://issues.redhat.com/browse/OSPRH-2451


kopecmartin:
I think that https://github.com/openstack-k8s-operators/tcib/pull/113 is a better approach that could help avoid any confusions about what is installed where as tempest-extras already clones other projects such as tempest-stress etc .. while the tempest-all doesn't clone any, it just installs openstack-tempest-all package

PR #119, opened 2024-01-17
Remove deprecated cinderlib from cinder volume

Akrog:
Cinderlib is being deprecated [1], so we should stop including it in the cinder-volume container.

[1]: https://review.opendev.org/q/topic:%22deprecate-cinderlib%22


ASBishop (approved):
/lgtm

raukadah (approved):

PR #120, opened 2024-01-18
[Tobiko] Get tobiko.conf and ssh keys from external_files dir

eduolivares:
The tobiko.conf file will be generated by ci-framework, which will be more flexible than creating it with crudini commands on the run_tobiko.sh script.

Another improvement is to obtain ssh key files generated by ci-framwork too. Those files can be shared between different tobiko containers running tests, so that they can connect to the VMs created by a previous tobiko container (the Openstack workloads created by tobiko are persistent).

Related: https://github.com/openstack-k8s-operators/ci-framework/pull/1044

### DIFF
--- a/container-images/tcib/base/os/tobiko/run_tobiko.sh
+++ b/container-images/tcib/base/os/tobiko/run_tobiko.sh
@@ -28,10 +28,12 @@ git clone https://opendev.org/x/tobiko
 pushd tobiko
 git checkout ${TOBIKO_VERSION}
 
-# obtain clouds.yaml and tobiko.conf from external_files directory
+# obtain clouds.yaml, id_ecdsa and tobiko.conf from external_files directory
 if [ ! -z ${USE_EXTERNAL_FILES} ]; then
     mkdir -p $TOBIKO_DIR/.config/openstack
-    cp $TOBIKO_DIR/external_files/clouds.yaml $HOME/.config/openstack/
+    cp $TOBIKO_DIR/external_files/clouds.yaml $TOBIKO_DIR/.config/openstack/
+    mkdir -p $TOBIKO_DIR/.ssh
+    cp $TOBIKO_DIR/external_files/id_ecdsa* $TOBIKO_DIR/.ssh/
     cp $TOBIKO_DIR/external_files/tobiko.conf .
 fi
 
--- a/container-images/tcib/base/os/tobiko/run_tobiko.sh
+++ b/container-images/tcib/base/os/tobiko/run_tobiko.sh
@@ -5,55 +5,35 @@ TOBIKO_DIR=/var/lib/tobiko
 # assert mandatory variables have been set
 [ -z ${TOBIKO_TESTENV} ] && echo "TOBIKO_TESTENV not set" && exit 1
 
-# obtain clouds.yaml from external_files directory
-if [ ! -z ${USE_EXTERNAL_FILES} ]; then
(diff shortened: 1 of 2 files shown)


container-images/tcib/base/os/tobiko/run_tobiko.sh
```diff
@@ -5,55 +5,37 @@ TOBIKO_DIR=/var/lib/tobiko
...
-crudini --set tobiko.conf testcase timeout ${TOBIKO_TESTCASE_TIMEOUT}
-crudini --set tobiko.conf testcase test_runner_timeout ${TOBIKO_TESTRUNNER_TIMEOUT}
-# ubuntu
-crudini --set tobiko.conf ubuntu interface_name ${TOBIKO_UBUNTU_INTERFACE_NAME}
-[ ! -z ${TOBIKO_UBUNTU_MINIMAL_IMAGE_URL} ] && crudini --set tobiko.conf ubuntu image_file ${TOBIKO_DIR}/.downloaded-images/ubuntu-minimal
-# keystone
-crudini --set tobiko.conf keystone interface ${TOBIKO_KEYSTONE_INTERFACE}
-[ ! -z ${TOBIKO_MANILA_SHARE_PROTOCOL} ] && crudini --set tobiko.conf manila share_protocol ${TOBIKO_MANILA_SHARE_PROTOCOL}
+# obtain clouds.yaml, id_ecdsa and tobiko.conf from external_files directory
+if [ ! -z ${USE_EXTERNAL_FILES} ]; then
```
marios, line 32: who/where are we setting that

marios (commented):
so i think this one depends-on the https://github.com/openstack-k8s-operators/ci-framework/pull/1044

i.e. do we need that one merged first?

that creates the tobiko.conf and this one consumes it right?

container-images/tcib/base/os/tobiko/run_tobiko.sh
eduolivares: it's here:
https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/tcib/base/os/tobiko/tobiko.yaml#L2

eduolivares:
> so i think this one depends-on the [openstack-k8s-operators/ci-framework#1044](https://github.com/openstack-k8s-operators/ci-framework/pull/1044)
> 
> i.e. do we need that one merged first?
> 
> that creates the tobiko.conf and this one consumes it right?
I'd say both depend on each other. Since we don't have any job using tobiko yet (that's my next task), it's safe to merge them in any order.

marios (approved):
ok thanks for checking in that case we don't need to order them

container-images/tcib/base/os/tobiko/run_tobiko.sh
```diff
@@ -5,55 +5,37 @@ TOBIKO_DIR=/var/lib/tobiko
...
+    mkdir -p $TOBIKO_DIR/.config/openstack
+    cp $TOBIKO_DIR/external_files/clouds.yaml $TOBIKO_DIR/.config/openstack/
+    mkdir -p $TOBIKO_DIR/.ssh
+    cp $TOBIKO_DIR/external_files/id_ecdsa* $TOBIKO_DIR/.ssh/
```
viroel, line 36: I was wondering if isn't possible to create a var for ssh keyname, so we don't hardcode this here.
Please consider an update in a follow up patch if you get a solution for this.
Thanks

viroel (approved):

PR #121, opened 2024-01-23
Remove python3-observabilityclient

arxcruz:
This package is not yet available as rpm, and it is blocking our promotion pipeline. It should be re-added once the package is ready.


rlandy:
recheck

rlandy:
recheck

PR #122, opened 2024-01-24
Refactor python3-observabilityclient package setup in tcib_packages

son-vyas:
This commit makes two significant changes to tcib pkg configuration:
1. Removes the 'python3-observabilityclient' package from the common configuration. This step is taken to streamline the package setup and tailor the dependencies more specifically to the needs of different architectures and operating systems.
2. Adds the 'python3-observabilityclient' package specifically for x86_64 and el8 architectures. This addition aligns with our strategy to optimize and enhance observability and monitoring capabilities on these architectures.

Additionally, a note is included to address future RHEL-specific package requirements, indicating that RHEL-specific packages can be added or uncommented once they are available.
This change prepares for a more tailored and efficient package management approach in our system configurations.

### DIFF
--- a/container-images/tcib/base/openstackclient/openstackclient.yaml
+++ b/container-images/tcib/base/openstackclient/openstackclient.yaml
@@ -16,11 +16,16 @@ tcib_packages:
   - python3-heatclient
   - python3-ironicclient
   - python3-manilaclient
-  - python3-observabilityclient
   - python3-octaviaclient
   - python3-aodhclient
   - bash-completion
   - iputils
+  #rhel: We can uncomment this when we have rhel specific pkgs
+  x86_64:
+  - python3-observabilityclient
+  el8:
+  - python3-observabilityclient
+
 tcib_user: cloud-admin
 tcib_workdir: /home/cloud-admin
 # If the container is run directly, start an interactive session


container-images/tcib/base/openstackclient/openstackclient.yaml
```diff
@@ -16,11 +16,16 @@ tcib_packages:
   - python3-heatclient
   - python3-ironicclient
   - python3-manilaclient
-  - python3-observabilityclient
   - python3-octaviaclient
   - python3-aodhclient
   - bash-completion
   - iputils
+  #rhel: We can uncomment this when we have rhel specific pkgs
```
marios, line 23: what will be uncommented here?

do you mean more like "we can add rhel specific pkgs to this list" 

[EDIT]: ok i see what you mean now. we can uncomment "rhel:" specifically to add a new 'rhel' key and package list

container-images/tcib/base/openstackclient/openstackclient.yaml
```diff
@@ -16,11 +16,16 @@ tcib_packages:
...
+  x86_64:
+  - python3-observabilityclient
+  el8:
```
marios, line 26: how/where are these 'architectures' decided. el8 is operating system specific and x86_64 is cpu architecture specific. how/where is this going to be used

arxcruz (commented):
Is this really work? I am checking the tcib python code, and I can't see anywhere differentiating

container-images/tcib/base/openstackclient/openstackclient.yaml
son-vyas: So, we had an issue with python3-observabilityclient pkg which we dont need in rhel now - https://redhat-internal.slack.com/archives/C04HN4L5XHV/p1706022079814729. So we want to follow the same pattern we followed with https://opendev.org/openstack/tripleo-common/src/branch/master/container-images/tcib/base/collectd/collectd.yaml#L50 to get this package excluded in rhel.

marios: ok thanks for the extra info.

as we discussed in yesterday's review call, lets keep only the one we need here (e.g. "el9" or "rhel9" or something similar) and then we will also need some further work to wire it up so that we get the right packages for rhel9

marios (changes requested):

container-images/tcib/base/openstackclient/openstackclient.yaml
```diff
@@ -16,11 +16,16 @@ tcib_packages:
   - python3-heatclient
   - python3-ironicclient
   - python3-manilaclient
-  - python3-observabilityclient
   - python3-octaviaclient
   - python3-aodhclient
   - bash-completion
   - iputils
+  #rhel: We can uncomment this when we have rhel specific pkgs
+  x86_64:
```
marios, line 24: (this one can be removed as discussed yesterday, we don't want/need to distinguish on cpu arch here)

paramite:
This package is already available in RDO all the way back down to Antelope release and also available in downstream for RHEL9. IMO the PR for observabilityclient should be reverted now.

paramite (changes requested):
IMO this is not needed any more as the package is available both for upstream and downstream.

container-images/tcib/base/openstackclient/openstackclient.yaml
mrunge: can I kindly ask for a consultation before you are actually removing packages that would affect other DFG?

container-images/tcib/base/openstackclient/openstackclient.yaml
```diff
@@ -16,11 +16,16 @@ tcib_packages:
   - python3-heatclient
   - python3-ironicclient
   - python3-manilaclient
-  - python3-observabilityclient
```
mrunge, line 19: We need python3-observabilityclient

container-images/tcib/base/openstackclient/openstackclient.yaml
SeanMooney: we should not have any el8 related content we will be building this for 9.4

container-images/tcib/base/openstackclient/openstackclient.yaml
```diff
@@ -16,11 +16,16 @@ tcib_packages:
...
   - python3-octaviaclient
   - python3-aodhclient
   - bash-completion
   - iputils
+  #rhel: We can uncomment this when we have rhel specific pkgs
+  x86_64:
+  - python3-observabilityclient
```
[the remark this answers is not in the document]
SeanMooney: does python3-observabilityclient provide an oepnstack client plugin? if not I'm not sure this is strictly in the scope fo this container image.

its not intended to have a set of generic debugging clients for example this container is not expect to have ssh or virsh for example to act as a way to debug the comptue nodes or even run must-gather or sosreports.

container-images/tcib/base/openstackclient/openstackclient.yaml
SeanMooney: we shoudl not have rhel sepfcic packages in this file this is intended to work on both centos and rhel.

and we do not expect people to edit the file so i dont think this comment is relevent

SeanMooney (changes requested):

container-images/tcib/base/openstackclient/openstackclient.yaml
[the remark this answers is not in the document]
mrunge: yes, it provides an openstack client plugin.

container-images/tcib/base/openstackclient/openstackclient.yaml
mrunge: observabilityclient is not built on el8, only for newer releases.

mrunge (commented):
I am not sold to this approach here.

PR #123, opened 2024-01-24
[Tempest][Tobiko] clouds.yaml mounted on the proper path

eduolivares:
When either a tempest or a tobiko container is run, the clouds.yaml file could be mounted directly on its proper path
`$HOME/.config/openstack/clouds.yaml` (new behavior implemented in ci-framework) or on the `external_files` directory (legacy behavior).

Related-PR: https://github.com/openstack-k8s-operators/ci-framework/pull/1057

[OSPRH-1670](https://issues.redhat.com//browse/OSPRH-1670)


eduolivares:
recheck

PR #124, opened 2024-01-24
[tempest-extras] Add neutron tempest plugin

jamepark4:
Add the neutron tempest plugin [1] to allow for more specific neutron coverage in network CI component pipeline.

[1] https://opendev.org/openstack/neutron-tempest-plugin


jamepark4:
Exists in tempest-all so this is unnecessary.

PR #129, opened 2024-02-05
[tempest] Fix parsing of TEMPESTCONF_{REMOVE,APPEND}

lpiwowar:
If there is a new line at the end of either TEMPESTCONF_REMOVE or TEMPESTCONF_APPEND then it might happen that the run_tempest.sh adds --remove/--append parameter to the tempest command with an empty value.

For example:
TEMPESTCONF_REMOVE="section.subsection False
" # <-- New line

leads to:
tempest run --remove section.subsection False --remove

Also, both --remove and --append expect '=' between section name and
value [1].


kopecmartin (approved):

kopecmartin (approved):
lgtm

kopecmartin:
recheck

lpiwowar:
recheck

PR #131, opened 2024-02-07
[Tobiko] Add tcpdump and prepare to run containers as root

eduolivares:
Some tobiko tests capture traffic on the tobiko container where tests are executed using tcpdump.

For that reason:
- tcpdump needs to be added to the list of packages that need to be installed on the tobiko image.
- the container will need to be executed with root permissions, so the run_tobiko.sh script needs to manage properly the mounted files and directories.

### DIFF
--- a/container-images/tcib/base/os/tobiko/run_tobiko.sh
+++ b/container-images/tcib/base/os/tobiko/run_tobiko.sh
@@ -1,9 +1,11 @@
 #!/bin/sh
 
+set -x
+
 TOBIKO_DIR=/var/lib/tobiko
 
 # assert mandatory variables have been set
-[ -z ${TOBIKO_TESTENV} ] && echo "TOBIKO_TESTENV not set" && exit 1
+[ -z "${TOBIKO_TESTENV}" ] && echo "TOBIKO_TESTENV not set" && exit 1
 
 # download Ubuntu minimal image used by the Tobiko scenario tests, if needed
 if [ ! -z ${TOBIKO_UBUNTU_MINIMAL_IMAGE_URL} ]; then
@@ -13,6 +15,7 @@ fi
 
 # set default values for the required variables
 TOBIKO_VERSION=${TOBIKO_VERSION:-master}
+TOBIKO_PRIVATE_KEY_FILE=${TOBIKO_PRIVATE_KEY_FILE:-id_ecdsa}
 
 # export OS_CLOUD variable
 [ ! -z ${TOBIKO_OS_CLOUD} ] && export OS_CLOUD=${TOBIKO_OS_CLOUD} || export OS_CLOUD=default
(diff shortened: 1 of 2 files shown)


container-images/tcib/base/os/tobiko/run_tobiko.sh
```diff
@@ -13,6 +15,7 @@ fi
 
 # set default values for the required variables
 TOBIKO_VERSION=${TOBIKO_VERSION:-master}
+TOBIKO_PRIVATE_KEY_FILE=${TOBIKO_PRIVATE_KEY_FILE:-id_ecdsa}
 
 # export OS_CLOUD variable
 [ ! -z ${TOBIKO_OS_CLOUD} ] && export OS_CLOUD=${TOBIKO_OS_CLOUD} || export OS_CLOUD=default
```
lpiwowar, line 21: This is just a small thing I noticed. It is not related to the PR but I wanted to at least mention it.
```bash
[ ! -z ${TOBIKO_NUM_PROCESSES} ] && export=TOX_NUM_PROCESSES=${TOBIKO_NUM_PROCESSES}
```
The '=' should not probably be here.

container-images/tcib/base/os/tobiko/run_tobiko.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -x
+
 TOBIKO_DIR=/var/lib/tobiko
 
 # assert mandatory variables have been set
```
lpiwowar, line 7: Not related to the PR. Maybe we can put quotation marks around the ${TOBIKO_TESTENV} here. For example, when the user wants to run a single test the TOBIKO_TESTENV variable will contain:
```
neutron -- tobiko/tests/scenario/neutron/test_qos.py::QoSNetworkTest::test_ping_dscp
```
and this will lead to:
```
tortuga.txt: line 4: [: syntax error: `--' unexpected
```

lpiwowar:
From the standpoint of the test-operator this change is backwards compatible and the tests that were working previously with the test-operator didn't break. So LGTM!
I'm going to test more thoroughly: 
```
neutron -- tobiko/tests/scenario/neutron/test_qos.py::QoSNetworkTest::test_ping_dscp
```

container-images/tcib/base/os/tobiko/run_tobiko.sh
eduolivares: good catch, thanks!

container-images/tcib/base/os/tobiko/run_tobiko.sh
eduolivares: good catch, thanks!

raukadah (approved):

PR #133, opened 2024-02-12
pin libvirt-libs due to RHEL-20609

SeanMooney:
This pins libvirt-libs to <9.10 to avoid
a unicode parseing bug that breaks nova.


SeanMooney:
this is not needed anymore

PR #134, opened 2024-02-12
pin libvirt-libs due to RHEL-20609

arxcruz:
Pinning only for centos stream


container-images/tcib/base/os/nova-base/nova-compute/nova-compute.yaml
```diff
@@ -3,6 +3,8 @@ tcib_actions:
 # this need to happen after installing nova-compute because the distgit does usermod to add libvirt/qemu groups
 - run: bash /usr/local/bin/uid_gid_manage nova
 - run: rm -f /etc/machine-id
+- run: >-
```
SeanMooney, line 6: from a compute perspective i have no issue with proceeding with this patch instead of mine

PR #138, opened 2024-02-22
Enable mounting of ssh keys to arbitrary folder

lpiwowar:
The test-operator requires the ability to specify arbitrary mount point
for the ssh keys. This patch introduces a new env variable called
TOBIKO_KEYS_FOLDER that contains a path to a folder containing
keys with ${TOBIKO_PRIVATE_KEY_FILE}* name.


lpiwowar:
I think the patch is ready. I just want to properly test it manually but from my quick testing it seems to be working.

eduolivares (approved):

eduolivares:
recheck

raukadah (approved):

PR #146, opened 2024-03-01
[Tobiko] Add support for separate log dirs

lpiwowar:
This change introduces TOBIKO_LOGS_DIR_NAME variable for the tobiko
image. This variable enables to specify directory name that should
contain the collected logs.


eduolivares (approved):

kopecmartin (approved):
lgtm

PR #147, opened 2024-03-04
Add displaying excluded and included tests

kstrenkova:


lpiwowar:
recheck

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -232,6 +232,16 @@ function run_rpm_tempest {
 function generate_test_results {
     pushd $TEMPEST_DIR
 
+    echo "Excluded tests"
+    if [ ! -z ${TEMPEST_EXCLUDE_LIST} ]; then
+        echo "$(cat ${TEMPEST_EXCLUDE_LIST})"
```
lpiwowar, line 237: Is the `echo` here really necessary?:)

kstrenkova: Thanks for review, it is not needed. My concern with missing end of line was caused by a small mistake in documentation.

slagle (approved):

PR #153, opened 2024-03-12
[Tobiko] Fix extraction of log dir name

lpiwowar:
Currently we look for the logs based on the name of testenv but we do not take into account the fact that the testenv can have the following format:

'functional -- tobiko/tests/functional/podified/test_topology.py'

This patch ensures that we take only the first part of the testenv name ('functional') into account when copying the logs.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1254


eduolivares (approved):

kopecmartin (approved):
lgtm

PR #154, opened 2024-03-14
Drop openstack-ironic-staging-drivers from ironic-condutor

raukadah:
https://issues.redhat.com/browse/OSPRH-5427 drops the openstack-ironic-staging-drivers from OSP-18 and package is removed from downstream dlrn builds, leading to failure in building ironic conductor container image.

Dropping this package from tcib fixes the issue.


amoralej (approved):
packaging wise, lgtm

PR #155, opened 2024-03-14
[Tempest-all] Add support for oc command

lpiwowar:
Whitebox-neutron-tempest-plugin requires oc command to be able to execute certain tests. This patch makes sure that oc is part of the tempest-all image.

### DIFF
--- a/container-images/tcib/base/os/tempest/run_tempest.sh
+++ b/container-images/tcib/base/os/tempest/run_tempest.sh
@@ -259,6 +259,10 @@ function generate_test_results {
 
 export OS_CLOUD=default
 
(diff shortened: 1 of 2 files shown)


container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -259,6 +259,10 @@ function generate_test_results {
 
 export OS_CLOUD=default
 
+if [ ! -z ${TEMPEST_OCP_CLIENT_TGZ} ]; then
```
lpiwowar, line 262: I'm wondering whether we want to install the oc cli by default in all tempest images or only in the tempest-all images. The installation is quite quick.

lpiwowar:
No longer needed.

PR #159, opened 2024-04-02
Fix private key permission

lpiwowar:
It seems that there is no direct way how to set 600 for a mounted secret using openshift. This patch creates a workoround that copies the mounted secret (private key) to the correct destination and sets the correct permissions for it.

### DIFF
--- a/container-images/tcib/base/os/tempest/run_tempest.sh
+++ b/container-images/tcib/base/os/tempest/run_tempest.sh
@@ -276,6 +276,16 @@ if [ ! -f ${TEMPEST_PATH}exclude.txt ] && [ -z ${TEMPEST_EXCLUDE_LIST} ]; then
     touch ${TEMPEST_PATH}exclude.txt
 fi
 
+# This workaround is required for the whitebox-neutron-tempest plugin. We need
+# to be able to specify 600 permissions for the id_ecdsa.
+if [ -f ${HOMEDIR}/id_ecdsa ]; then
+    mkdir -p ${HOMEDIR}/.ssh
+    cp ${HOMEDIR}/id_ecdsa ${HOMEDIR}/.ssh/id_ecdsa
+    sudo chmod 700 ${HOMEDIR}/.ssh
+    sudo chmod 600 ${HOMEDIR}/.ssh/id_ecdsa
+    sudo chown -R tempest:tempest ${HOMEDIR}/.ssh
+fi
+
 if [ -z $TEMPEST_EXTERNAL_PLUGIN_GIT_URL ]; then
     run_rpm_tempest
 else


container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -276,6 +276,16 @@ if [ ! -f ${TEMPEST_PATH}exclude.txt ] && [ -z ${TEMPEST_EXCLUDE_LIST} ]; then
     touch ${TEMPEST_PATH}exclude.txt
 fi
 
+# This workaround is required for the whitebox-neutron-tempest plugin. We need
+# to be able to specify 600 permissions for the id_ecdsa.
+if [ -f ${HOMEDIR}/id_ecdsa ]; then
```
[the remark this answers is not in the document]
lpiwowar: Yes, the previous behavior of the test-operator was to mount the key under `/var/lib/tempest/.ssh` but this will be changed with these two PRs:

- https://github.com/openstack-k8s-operators/tcib/pull/159
- https://github.com/openstack-k8s-operators/test-operator/pull/69

kopecmartin (approved):
lgtm

kopecmartin:
recheck

container-images/tcib/base/os/tempest/run_tempest.sh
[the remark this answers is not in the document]
lpiwowar: I guess that from the previous comments it can be a little bit confusing but it is actually the other way around. [The patch in the test-operator](https://github.com/openstack-k8s-operators/test-operator/pull/69/files) depends on this change. 

Merging this change before the one in the test-operator should not have any influence on the run_tempest.sh thanks to this condition:
```
if [ -f ${HOMEDIR}/id_ecdsa ]; then
```

m-blue-m (changes requested):

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -276,6 +276,16 @@ if [ ! -f ${TEMPEST_PATH}exclude.txt ] && [ -z ${TEMPEST_EXCLUDE_LIST} ]; then
...
+    mkdir ${HOMEDIR}/.ssh
+    cp ${HOMEDIR}/id_ecdsa ${HOMEDIR}/.ssh/id_ecdsa
+    chmod 700 ${HOMEDIR}/.ssh
+    chmod 600 ${HOMEDIR}/.ssh/id_ecdsa
+    chown -R tempest:tempest ${HOMEDIR}/.ssh/id_ecdsa
```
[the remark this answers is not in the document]
lpiwowar: Thank you:). You're right. I updated the patch.

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -276,6 +276,16 @@ if [ ! -f ${TEMPEST_PATH}exclude.txt ] && [ -z ${TEMPEST_EXCLUDE_LIST} ]; then
     touch ${TEMPEST_PATH}exclude.txt
 fi
 
+# This workaround is required for the whitebox-neutron-tempest plugin. We need
+# to be able to specify 600 permissions for the id_ecdsa.
+if [ -f ${HOMEDIR}/id_ecdsa ]; then
+    mkdir ${HOMEDIR}/.ssh
+    cp ${HOMEDIR}/id_ecdsa ${HOMEDIR}/.ssh/id_ecdsa
```
m-blue-m, line 283: Thank you for the changes, looks good.
Only on my setup deployed before these changes, it seem 'id_ecdsa' was root owned, so maybe it would need 'sudo cp ...' too.
`
sh-5.1$ ls -lh /var/lib/tempest/.ssh/id_ecdsa
-rw-r--r--. 1 root tempest 512 Apr  4 10:12 /var/lib/tempest/.ssh/id_ecdsa
`

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -276,6 +276,16 @@ if [ ! -f ${TEMPEST_PATH}exclude.txt ] && [ -z ${TEMPEST_EXCLUDE_LIST} ]; then
     touch ${TEMPEST_PATH}exclude.txt
 fi
 
+# This workaround is required for the whitebox-neutron-tempest plugin. We need
+# to be able to specify 600 permissions for the id_ecdsa.
+if [ -f ${HOMEDIR}/id_ecdsa ]; then
+    mkdir ${HOMEDIR}/.ssh
```
m-blue-m, line 282: optional nit: sometimes ssh commands create .ssh, it is possible that in the future it would be created earlier, so "if dir doesn't exist, then create dir" condition can prevent non-zero exit status

m-blue-m (changes requested):

container-images/tcib/base/os/tempest/run_tempest.sh
lpiwowar: I do not think we are going to run ssh command in this script but I don't see a reason why we couldn't add  `mkdir -p`:)

container-images/tcib/base/os/tempest/run_tempest.sh
lpiwowar: It should not be a problem, shouldn't it? The file is owned by the tempest group with read access.

m-blue-m: Right, that would work too if mkdir runs as tempest user.
All looks good to me, thanks.

raukadah:
recheck

marios (approved):
ok thanks for clarification, cant spot something 

last ci fail seems unrelated (timeout on https://logserver.rdoproject.org/59/159/a0e73f9d6c89832f6aad3d26e33073a7e6d77728/github-check/tcib-podified-multinode-edpm-deployment-crc/db4e5b0/controller/ci-framework-data/tests/test_operator/stestr_results.html

Timed out waiting for 192.168.122.225 to become reachable

raukadah (approved):

raukadah:
recheck after tempest fix

kopecmartin:
recheck failure looks random (judging by the build history)

raukadah:
recheck

fyanac:
recheck

raukadah:
recheck tempest fix merged

PR #161, opened 2024-04-04
Add ipmitool to openstackclient base image

lmiccini:
Having ipmitool in the client image could be useful for troubleshooting purposes as well as allowing us to re-use the client image for things like Instance HA, avoiding the burden of having to maintain another dedicated image.


raukadah:
recheck

raukadah (approved):

PR #165, opened 2024-04-16
[Tobiko] Download tobiko repo, oc and ubuntu image during image creation

eduolivares:
The tobiko repo, the openshift client (oc) and the ubuntu image
were downloaded when the tobiko container started running, before this
patch.
After this patch, those operations will be performed during the image
creation, so when the container starts running tobiko, oc and ubuntu
image will already be available.

[OSPRH-6347](https://issues.redhat.com//browse/OSPRH-6347)

### DIFF
--- a/container-images/tcib/base/tobiko/run_tobiko.sh
+++ b/container-images/tcib/base/tobiko/run_tobiko.sh
@@ -7,16 +7,9 @@ TOBIKO_DIR=/var/lib/tobiko
 # assert mandatory variables have been set
 [ -z "${TOBIKO_TESTENV}" ] && echo "TOBIKO_TESTENV not set" && exit 1
 
-# download Ubuntu minimal image used by the Tobiko scenario tests, if needed
-if [ ! -z ${TOBIKO_UBUNTU_MINIMAL_IMAGE_URL} ]; then
-    mkdir -p ${TOBIKO_DIR}/.downloaded-images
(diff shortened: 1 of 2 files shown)


container-images/tcib/base/tobiko/run_tobiko.sh
```diff
@@ -30,8 +29,9 @@ TOBIKO_LOGS_DIR_NAME=${TOBIKO_LOGS_DIR_NAME:-"tobiko"}
 [ ! -z ${TOBIKO_NUM_PROCESSES} ] && export TOX_NUM_PROCESSES=${TOBIKO_NUM_PROCESSES}
 
 pushd ${TOBIKO_DIR}
-git clone https://opendev.org/x/tobiko
+sudo chown tobiko:tobiko -R tobiko
 pushd tobiko
+[ ! -z ${TOBIKO_UPDATE_REPO} ] && git pull --rebase
```
lpiwowar, line 34: We probably need an option for this in the Tobiko CR, right?

eduolivares: yes, but it will be disabled by default, so no rush

eduolivares:
@lpiwowar @tosky apart from what we said about cloning tobiko and downloading oc, I moved the download of the ubuntu image to the tobiko image creation as part of this PR:
https://github.com/openstack-k8s-operators/tcib/compare/2ffb51588c4f3a6ede471fb222641784eb10eef3..15b18f69ac928ec1bad4eb762c7a60789d72a913

lpiwowar (approved):
Looks good to me! :+1:

raukadah (approved):

PR #169, opened 2024-05-02
Download advanced image only if glance image don't exist

karelyatin:
We do not need to re download the image on tempest rerun if image already created.


lpiwowar (approved):
LGTM, thank you!

kopecmartin (approved):
lgtm, thanks for noticing this!

PR #170, opened 2024-05-07
[Tempest] Make TEMPEST_NEUTRON_IMAGE more generic 

lpiwowar:
The TEMPEST_NEUTRON_IMAGE_* variables were tailored specifically to the
neutron-tempest-plugin. Those variables did not allow to specify things
like flavor, ID or name of the image that should be uploaded to
OpenStack.

This patch introduces a new set of variables called
TEMPEST_EXTRA_IMAGES_* that can be used to specify a list of images
that should be uploaded to OpenStack. These variables accepted
a comma-separated list of values.

### DIFF
--- a/container-images/tcib/base/os/tempest/run_tempest.sh
+++ b/container-images/tcib/base/os/tempest/run_tempest.sh
@@ -136,12 +136,40 @@ TEMPEST_EXTERNAL_PLUGIN_CHANGE_URL="${TEMPEST_EXTERNAL_PLUGIN_CHANGE_URL:-}"
 TEMPEST_EXTERNAL_PLUGIN_REFSPEC="${TEMPEST_EXTERNAL_PLUGIN_REFSPEC:-}"
 TEMPEST_EXTERNAL_PLUGIN_DIR=/var/lib/tempest/external-plugins
 
+TEMPEST_EXTRA_IMAGES_URL="${TEMPEST_EXTRA_IMAGES_URL:-}"
+TEMPEST_EXTRA_IMAGES_DISK_FORMAT="${TEMPEST_EXTRA_IMAGES_DISK_FORMAT:-}"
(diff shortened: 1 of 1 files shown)


container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -290,8 +343,8 @@ function generate_test_results {
 
 export OS_CLOUD=default
 
-if [[ ! -z ${TEMPEST_NEUTRON_IMAGE_URL} ]]; then
-    generate_extra_tempest_configuration
+if [[ ! ${#TEMPEST_EXTRA_IMAGES_NAME[@]} -eq 0 ]]; then
```
kopecmartin, line 346: what if TEMPEST_EXTRA_IMAGES_NAME is not set, does it work?

I tried this:
```
$ cat sh.sh 
TEMPEST_EXTRA_IMAGES_NAME="${TEMPEST_EXTRA_IMAGES_NAME:-}"

if [[ ! ${#TEMPEST_EXTRA_IMAGES_NAME[@]} -eq 0 ]]; then
    echo "image handling"
fi
```

and it printed the echo - it got inside the if and it shouldn't
```
$ sh sh.sh 
image handling
```

lpiwowar: In the example, there is conversion to an array missing (line 143). This works:

```
TEMPEST_EXTRA_IMAGES_NAME="${TEMPEST_EXTRA_IMAGES_NAME:-}"
read -ra TEMPEST_EXTRA_IMAGES_NAME <<< $TEMPEST_EXTRA_IMAGES_NAME

if [[ ! ${#TEMPEST_EXTRA_IMAGES_NAME[@]} -eq 0 ]]; then
    echo "image handling"
fi
```

This works ^^.

kopecmartin (approved):
lgtm

PR #172, opened 2024-05-27
glance-api: install python3-boto3 as to support s3

CyrilRoelandteNovance:
This dependency is required for users trying to use the s3 backend.


konan-abhi (commented):
this looks good and straightforward , but are we going to support s3 for glance?

PR #174, opened 2024-06-03
[tempest] TEMPEST_PARALLEL shouldn't have default

kopecmartin:
TEMPEST_PARALLEL variable should not have a default value set. Tempest has a default value, setting another default value in, basically, a wrapper introduces unnecessary edge cases, e.g. if someone sets TEMPEST_SERIAL and doesn't notice that TEMPEST_PARALLEL has a default value, it will result in setting both (--serial and --parallel) and Tempest will fail on that.


lpiwowar (approved):
LGTM! :+1:

kopecmartin:
recheck

raukadah (approved):

marios (approved):
lgtm

PR #175, opened 2024-06-06
Install extra rpms

lpiwowar:
This patch introduces TEMPEST_EXTRA_RPMS variable that accepts a comma-separated list of URLs that point to RPMs that should be installed before the execution of tempest.


container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -313,6 +316,11 @@ function run_rpm_tempest {
     tempest init openshift
     pushd $TEMPEST_DIR
 
+    # Install additional plugins from .rpms plus their dependencies
+    for rpm_package in "${TEMPEST_EXTRA_RPMS[@]}"; do
+        sudo rpm -ivh $rpm_package
```
lpiwowar, line 321: I'm thinking about two things here:
- Should I add `--force` to overwrite files if they already exist (in case we want to install a newer version of an rpm)
- Is it ok to keep the installation of the dependencies up to the user? Do we want to enable the user to specify additional repositories from which the dependencies can be installed?

kopecmartin: > * Should I add `--force` to overwrite files if they already exist (in case we want to install a newer version of an rpm)

hmm, newer package of something that is already in? Sounds like they should wait until the image gets rebuild with the newer deps, or they can manually ssh in .. this sounds to me like a very small edge case.

> * Is it ok to keep the installation of the dependencies up to the user? Do we want to enable the user to specify additional repositories from which the dependencies can be installed?

why wouldn't it be ok? i wouldn't add more functionalities, it's still a test-operator, not an installer. If very specific repos and packages are needed, users still can write their playbook to ssh to the pod and do what they need to do, before running the tests.

kopecmartin (approved):
lgtm

container-images/tcib/base/os/tempest/run_tempest.sh
lpiwowar: Thanks @kopecmartin. I think those are valid points. I'm also curious what Luigi (@tosky) thinks since he "ordered" the change.

tosky: So for this special case it would be up to each user to find and install the dependencies.
But I don't think this is going to work, or at least it works if the dependencies are in the right order.
For example, if you pass A.rpm and B.rpm and B.rpm depends on A.rpm, the order matters.
Why not just scrap the loop and just use something like:

sudo dnf install "${TEMPEST_EXTRA_RPMS[@]}"

?

lpiwowar: I updated the code. Thanks! I did not know you can pass all the rpms to dnf at once.

kopecmartin (approved):
thanks

lpiwowar:
recheck

lpiwowar:
recheck

Seems like the last results from zuul were not successfully linked to GitHub (results stuck in Pending state)

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -313,6 +316,9 @@ function run_rpm_tempest {
     tempest init openshift
     pushd $TEMPEST_DIR
 
+    # Install additional plugins from .rpms plus their dependencies
+    [ ${#TEMPEST_EXTRA_RPMS[@]} -ne 0 ] && sudo dnf install -y ${TEMPEST_EXTRA_RPMS[@]}
```
marios, line 320: i was unsure of that bash syntax but it is correct 

${#TEMPEST_EXTRA_RPMS[@]}

the # gives the number of elements

marios (approved):

PR #177, opened 2024-06-12
Init openstack-rally image

kopecmartin:
This prepares openstack-rally image so that it's ready when someone wants to run rally via podman or openshift.


container-images/tcib/base/os/rally/rally.yaml
```diff
@@ -0,0 +1,21 @@
...
+- run: chown -R rally.rally /var/lib/rally
+- run: touch /var/lib/kolla/config_files/config.json
+- run: cp /usr/share/tcib/container-images/tcib/base/os/rally/run_rally.sh /var/lib/rally/run_rally.sh
+- run: chmod +x /var/lib/rally/run_rally.sh
+
+tcib_entrypoint: /var/lib/rally/run_rally.sh
+
+tcib_packages:
+  common:
+  - openstack-rally
```
karelyatin, line 19: don't we need to include openstack-rally-plugins[1] rpm?

[1] https://github.com/rdo-packages/rally-openstack-distgit/blob/rpm-master/openstack-rally-plugins.spec#L15

raukadah:
@kopecmartin Please rebase your pr from tcib main branch. Horizontest issue might get fixed.

container-images/tcib/base/os/rally/rally.yaml
kopecmartin: i honestly have no idea, I've never used rally .. my only goal is to have that image ready once someone needs it and if someone does need it, they can build the rest of the script (or the list of the dependencies) according their needs

karelyatin: Ok so no one requested it already? afaik that's used together considering previous OSP releases so asked about that. The original request that resulted into this image should have the required information.

raukadah:
recheck 

tobiko issue fixed!

container-images/kolla/base/uid_gid_manage.sh
```diff
@@ -65,6 +65,7 @@ _SUPPORTED_USERS['ovn-bgp']='ovn-bgp 42486 42486 /var/lib/ovn-bgp kolla'
 _SUPPORTED_USERS['placement']='placement 42482 42482 /var/lib/placement kolla'
 _SUPPORTED_USERS['qemu']='qemu 107 107'
 _SUPPORTED_USERS['rabbitmq']='rabbitmq 42439 42439 /var/lib/rabbitmq kolla'
+_SUPPORTED_USERS['rally']='rally 42480 42480 /var/lib/rally kolla'
```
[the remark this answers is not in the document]
raukadah: It is the userid and groupid which we need to specify while creating the user with in the container. It has a sequence starting with 424. I am not sure about last two digits. It might be random and unique. May be we can use the random and different id for rally.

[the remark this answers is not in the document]
karelyatin: From background perspective it used to be coming from kolla side and there were some constraints related to rpm as well but i don't recall. 

Although it's dropped in kolla but i think can use that value here and keep seperate from tempest:- https://github.com/openstack/kolla/blob/428f37124986019bedc897469f9b1f3470b62082/kolla/common/users.py#L160-L162

container-images/tcib/base/os/rally/rally.yaml
karelyatin: @eduolivares can you confirm this so it can get included here.

eduolivares: There are two sets of downstream jobs using rally now.

The neutron jobs running rally use the rally infrared plugin. That plugin allows to install rally using three different methods: git, pip and rpm.
https://github.com/redhat-openstack/infrared/tree/master/plugins/rally/tasks/setup
The default method is pip:
https://github.com/redhat-openstack/infrared/blob/master/plugins/rally/plugin.spec#L36
If I'm not wrong, installing extra plugins is always done with git:
https://github.com/redhat-openstack/infrared/blob/master/plugins/rally/tasks/setup/rpm/main.yml#L35
And, actually, the neutron job does not add any of these extra plugins. This is from a downstream job:
```
2024-06-20 23:17:22.673 | TASK [Clone Rally-Plugins git repo] ********************************************
2024-06-20 23:17:22.674 | task path: /home/rhos-ci/jenkins/workspace/DFG-network-neutron-17.1_director-rhel-virthost-3cont_2comp-ipv6-all-geneve-dvr-stf-rally/infrared/plugins/rally/tasks/setup/pip/main.yml:34
2024-06-20 23:17:22.676 | Thursday 20 June 2024  23:17:21 +0000 (0:00:29.075)       0:01:00.118 ********* 
2024-06-20 23:17:22.679 | skipping: [undercloud-0] => {
2024-06-20 23:17:22.681 |     "changed": false,
2024-06-20 23:17:22.683 |     "skip_reason": "Conditional result was False"
2024-06-20 23:17:22.684 | }
```

Regarding the other downstream jobs using rally, the perfscale jobs, it seems they install browbeat from git at this moment. If it can be installed from RPMs, that would be much better, I guess.


So, sorry for the long comment, but, in conclussion, if plugins (at least browbeat) can be installed from RPMs, that would be nice.

lpiwowar (approved):
LGTM :+1:

container-images/tcib/base/os/rally/rally.yaml
karelyatin: @eduolivares thx for the response what i meant was specific to rally-openstack inclusion in the image and looking infrared, as defaults is pip, i see downstream rally jobs using that only and  have rally-openstack installed always as per https://github.com/redhat-openstack/infrared/blob/master/plugins/rally/tasks/setup/pip/main.yml#L17-L32, so would have been good to include the rpm for it in the image as well, but since it's already merged can be a follow up when needed :)

PR #180, opened 2024-06-19
Don't run Tempest if discover failed

kopecmartin:
Tempest run requires a tempest.conf, without it, it does not make sense to run Tempest. If discover-tempest-config fails for any reason, we should not run tempest to save execution time.

### DIFF
--- a/container-images/tcib/base/os/tempest/run_tempest.sh
+++ b/container-images/tcib/base/os/tempest/run_tempest.sh
@@ -300,8 +300,8 @@ function run_git_tempest {
     tempest init openshift
     pushd $TEMPEST_DIR
 
-    discover-tempest-config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES}
-    tempest run ${TEMPEST_ARGS}
+    discover-tempest-config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
+    && tempest run ${TEMPEST_ARGS}
     RETURN_VALUE=$?
 
     deactivate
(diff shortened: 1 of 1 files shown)


container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -300,8 +300,8 @@ function run_git_tempest {
     tempest init openshift
     pushd $TEMPEST_DIR
 
-    discover-tempest-config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES}
-    tempest run ${TEMPEST_ARGS}
+    discover-tempest-config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
+    && tempest run ${TEMPEST_ARGS}
```
kopecmartin, line 304: technically it could be like this:
`
command $$
command
`
&& doesn't require \ , bash knows another command follows .. but it looked strange to me, so I chose this style to really highlight that the second command depends on the first one being successful :
`
command \
&& command
`

lpiwowar (approved):
LGTM! :+1:

lpiwowar:
recheck
```
    nodenetworkconfigurationpolicy.nmstate.io/enp6s0-crc created
    timeout 240s bash -c "while ! (oc wait nncp -l osp/interface=enp6s0 --for jsonpath='{.status.conditions[0].reason}'=SuccessfullyConfigured); do sleep 10; done"
    error: timed out waiting for the condition on nodenetworkconfigurationpolicies/enp6s0-crc
    error: timed out waiting for the condition on nodenetworkconfigurationpolicies/enp6s0-crc
    error: timed out waiting for the condition on nodenetworkconfigurationpolicies/enp6s0-crc
    error: timed out waiting for the condition on nodenetworkconfigurationpolicies/enp6s0-crc
    error: timed out waiting for the condition on nodenetworkconfigurationpolicies/enp6s0-crc
    error: timed out waiting for the condition on nodenetworkconfigurationpolicies/enp6s0-crc
```
Suspicious timeout.

kopecmartin:
recheck

PR #183, opened 2024-07-01
Move setting tobiko dir owner to the last action

eduolivares:
After the owner of the tobiko home directory (/var/lib/tobiko) was set, some other actions like creating a subdir were done using the root user.
This patch moves the action that sets the owner of that directory and its subdirectories to the end of the list of actions.


kopecmartin (approved):
lgtm

rlandy (approved):
/approved

PR #185, opened 2024-07-11
Remove parallel option

eshulman2:
parallel option is default behavior of tempest and there is no need to specify it.

### DIFF
--- a/container-images/tcib/base/os/tempest/run_tempest.sh
+++ b/container-images/tcib/base/os/tempest/run_tempest.sh
@@ -43,7 +43,7 @@
 # responsible for the tempest tests execution.
 #
 # Supported boolean arguments:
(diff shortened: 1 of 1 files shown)


raukadah:
@kopecmartin @lpiwowar Hello, Please have a look at this pr, thank you :-)

eshulman2:
@lpiwowar maybe we should close it for now and go back to the parallel thing when we have time

lpiwowar (commented):
@eshulman2 I think we can go forward with this PR but it should be followed with PR for test-operator that removes the parallel option from the Tempest CR as well. Once I see that there is PR for the test-operator as well then I would give +1. 

It's up to you if you want to invest time in this right now. If not then I think it is ok to leave it for now:).

rabi (approved):
/lgtm

PR #187, opened 2024-07-17
Add required variables for horizon.conf

HanzJas:
Add required variables/values for horizon.conf which is used as a source of default values for Horizon integration tests.


ashu-011 (approved):
LGTM!

raukadah (commented):
/approve

PR #191, opened 2024-07-31
Drop etcd from tcib

raukadah:


lhh:
This is fine. My understanding is that there is no operator for this container at the moment; if we need one later, we can put it back.

viroel (approved):

rlandy (approved):
/lgtm

PR #192, opened 2024-07-31
Fix Horizontest pod exit

ashu-011:
We need to make sure Horizontest pod exits correctly , based on the exit value the pod either fails or succeeds . This patch fixed that part


lpiwowar (approved):
LGTM! :+1:

PR #193, opened 2024-08-01
Add Tobiko debug mode for test-operator

kstrenkova:
Similar to debug mode for Tempest, this patch introduces new option for test-operator that keeps the pod running after a successful run or even after a failure has occured. It resolves the issue that oc rsh session is killed once the execution of the run_tobiko.sh file finishes.


lpiwowar (approved):
LGTM! :+1: Thank you!

PR #196, opened 2024-08-08
Wait for image to be active after creation

lkuchlan:
We encountered Tempest failures due to the image being in the importing process.
This patch ensures the image is active when the Tempest tests are executed.

Jira: https://issues.redhat.com/browse/OSPRH-9284

### DIFF
--- a/container-images/tcib/base/os/tempest/run_tempest.sh
+++ b/container-images/tcib/base/os/tempest/run_tempest.sh
@@ -95,6 +95,7 @@ CONCURRENCY="${CONCURRENCY:-}"
 TEMPESTCONF_ARGS=""
 TEMPEST_ARGS=""
 TEMPEST_DEBUG_MODE="${TEMPEST_DEBUG_MODE:-false}"
+TEMPEST_IMAGE_CREATE_TIMEOUT="${TEMPEST_IMAGE_CREATE_TIMEOUT:-300}"
 
 function catch_error_if_debug {
     echo "File run_tempest.sh has run into an error!"
@@ -217,6 +218,10 @@ if [ -n "$CONCURRENCY" ] && [ -z ${TEMPEST_CONCURRENCY} ]; then
     TEMPEST_ARGS+="--concurrency ${CONCURRENCY} "
 fi
 
+function get_image_status {
+    openstack image show $IMAGE_ID -f value -c status
+}
+
 function upload_extra_images {
     for image_index in "${!TEMPEST_EXTRA_IMAGES_NAME[@]}"; do
         if ! openstack image show ${TEMPEST_EXTRA_IMAGES_NAME[image_index]}; then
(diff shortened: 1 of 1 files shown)


container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -241,7 +245,14 @@ function upload_extra_images {
                 image_create_params+=(--container-format ${TEMPEST_EXTRA_IMAGES_CONTAINER_FORMAT[image_index]})
 
             image_create_params+=(--public ${TEMPEST_EXTRA_IMAGES_NAME[image_index]})
-            openstack image create --import ${image_create_params[@]}
+            IMAGE_ID=$(openstack image create --import ${image_create_params[@]} -f value -c id)
+            STATUS=$(get_image_status)
+            while [ "$STATUS" != "active" ]; do
```
lpiwowar, line 250: Do we really want this to be looping in while? What if there is an issue with uploading the image and the job will be stuck until it times out. We should fail as quickly as possible I think.

What about trying to create the image 5 times in a for loop and if it succeeds then just lets break out of the for loop or something similar? And maybe we should fail the whole script when the image creation fails.

lpiwowar (changes requested):
This change is really needed! 

I've just opened a discussion with one comment. Let's see what other think.

container-images/tcib/base/os/tempest/run_tempest.sh
lkuchlan: I agree that we should exit the loop, but I'm unsure what a reasonable time frame is for the image to reach the active state, as there are several factors that can affect this. Let's set a relatively long timeout, around 5 minutes, and then break out of the loop. Since these images are for scenario tests, a failure in image creation would cause all scenario tests to fail, so I'm uncertain whether we should fail the entire script in this case. WDYT ?

lpiwowar: >  Let's set a relatively long timeout, around 5 minutes, and then break out of the loop.

Yes, true. I think 5 minutes should be relatively long enough. And it is better to have some timeout than none.

>  Since these images are for scenario tests, a failure in image creation would cause all scenario tests to fail, so I'm uncertain whether we should fail the entire script in this case.

The think is that we would be getting false negatives because the failure of the tests would not be caused by the problem in openstack but because of an error in the preparation of the test environment. I think we should keep these failures to minimum. So I would vote for ending the script with an error. We've already done something similar here [1] (when discover-tempest-config fails).

[1] [here](https://github.com/openstack-k8s-operators/tcib/blob/250504bf85b3d0e9a6c332521684566f4f2ad7a7/container-images/tcib/base/os/tempest/run_tempest.sh#L322)

lpiwowar (commented):
Looks good to me! :+1: 
Just one small comment.

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -95,6 +95,7 @@ CONCURRENCY="${CONCURRENCY:-}"
 TEMPESTCONF_ARGS=""
 TEMPEST_ARGS=""
 TEMPEST_DEBUG_MODE="${TEMPEST_DEBUG_MODE:-false}"
+IMAGE_TIMEOUT=300
```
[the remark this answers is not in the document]
lkuchlan: Right, Done!

lpiwowar (approved):
Looks good to me! :+1:

raukadah (approved):

PR #197, opened 2024-08-08
[Tempest] Remove cirros image from logs

lpiwowar:
Up until now we were collecting cirros image that is downloaded during the discover-tempest-config phase into the logs folder. The cirros image is not needed artifact and in some edge cases it might happen that the cirros file is too big for it to be stored in the external PV.

This patch removes the cirros image before the logs are collected.


rdiazcam:
lgtm

son-vyas:
just a small nit, like have we tested this change like removing the image not breaking any other stuffs? like just a sanity check would be a good thing to have

lpiwowar:
@son-vyas I've tested the change locally and @rdiazcam tested the change as well in his environment. How do you think this change will break the tests execution when we are deleting the image after the tests were executed?

marios (approved):

kopecmartin (approved):
lgtm, thanks

PR #198, opened 2024-08-09
Clean leftover images

ashu-011:
This patch temporarily handles the cleanup of any leftover images from tempest until tempest clean-up is implemented


container-images/tcib/base/os/horizontest/run_horizontest.sh
```diff
@@ -25,6 +25,13 @@ SELENIUM_IMPLICIT_WAIT=30
 [[ -z ${REPO_URL} ]] && REPO_URL="https://review.opendev.org/openstack/horizon"
 [[ -z ${HORIZON_REPO_BRANCH} ]] && HORIZON_REPO_BRANCH="master"
 
+#This function is temporarily added until tempest cleanup is implemented
+function clean_leftover_images {
+    if openstack image list -c Name -f value --os-cloud default ; then
```
lpiwowar, line 30: I'm just thinking. Is the if-statement needed?

ashu-011: Not a need here . let me remove it

PR #207, opened 2024-09-05
Missing defaults in ansible-tests

kopecmartin:
The run_ansible.sh was missing default values for couple of parameters potentially causing "unbound variable" error.


kopecmartin:
recheck

lpiwowar (changes requested):

lpiwowar (approved):
Looks good to me! :+1:

PR #208, opened 2024-09-09
Cleanup tempest related resources after testing

kopecmartin:
Let's run tempest cleanup after the tempest run, unless debug mode is enabled, to delete any leftover resources (from the tempest run) so that they won't disrupt any further testing - testing frameworks executed after tempest.


container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -319,9 +319,16 @@ function run_git_tempest {
...
 
     discover-tempest-config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
+    && tempest cleanup --init-saved-state || true \
     && tempest run ${TEMPEST_ARGS}
     RETURN_VALUE=$?
 
+    # Run tempest cleanup to delete any leftover resources
+    # when not in debug mode
+    if [ ${TEMPEST_DEBUG_MODE} == false ]; then
+        tempest cleanup || true
```
lpiwowar, line 329: I think it is ok to have the command without `|| true` here. The script will continue execution even when the command fails. The failure of the whole pod is defined by RETURN_VALUE=$?.

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -341,9 +348,16 @@ function run_rpm_tempest {
...
```
lpiwowar, line 358: ditto

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -319,9 +319,16 @@ function run_git_tempest {
     pushd $TEMPEST_DIR
 
     discover-tempest-config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
+    && tempest cleanup --init-saved-state || true \
```
lpiwowar, line 322: I think this should be moved before `upload_extra_images` function call and call of discover-tempest-config as it creates some resources sometimes as well (e.g., upload of cirros image).

lpiwowar (changes requested):
This will run tempest cleanup at the end of execution of each test pod. This means that we are going to be downloading and uploading images each time a new pod is spawned (after the necessary fix I proposed). 

Meaning, if there are 3 workflow steps we are going to be downloading and deleting image 3 times. Considering that some of the images are big it would be waste of time (we are talking 15 mins per download).

Here are two things that we could do:
- Modify this patch so that the cleanup behavior can be turned on / off through env variable (default off)
- Allow setting this parameter through Tempest CR and test_operator role.

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -340,10 +353,22 @@ function run_rpm_tempest {
...
+
     eval discover-tempest-config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
     && tempest run ${TEMPEST_ARGS}
     RETURN_VALUE=$?
 
+    # Run tempest cleanup to delete any leftover resources
+    # when not in debug mode
+    if [ ${TEMPEST_DEBUG_MODE} == false && ${TEMPEST_CLEANUP} == true ]; then
+        tempest cleanup
+    fi
```
lpiwowar, line 370: If you want to use `&&` and `[` then the correct syntax is this:
```
[ ${TEMPEST_DEBUG_MODE} = false] && [ ${TEMPEST_CLEANUP} = true ]
```
otherwise you should use `[[`:
```
[[ ${TEMPEST_DEBUG_MODE} == false] && [ ${TEMPEST_CLEANUP} == true ]]
```

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -340,10 +353,22 @@ function run_rpm_tempest {
     # List Tempest packages
     rpm -qa | grep tempest
 
+    tempest cleanup --init-saved-state
+
+    if [[ ! ${#TEMPEST_EXTRA_IMAGES_NAME[@]} -eq 0 ]]; then
+        upload_extra_images
+    fi
```
lpiwowar, line 360: I know this code already existed here and this change just moves it around. But since there are other changes required we can simplify [1] this as:
```
# No if-statement
upload_extra_images
```

[1] https://github.com/openstack-k8s-operators/tcib/pull/208/files#diff-191bc7762340f2539b035115667e34a49916229d47d776aca4a2c457f694e35cR228

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -318,10 +319,22 @@ function run_git_tempest {
     tempest init openshift
     pushd $TEMPEST_DIR
 
+    tempest cleanup --init-saved-state
+
+    if [[ ! ${#TEMPEST_EXTRA_IMAGES_NAME[@]} -eq 0 ]]; then
+        upload_extra_images
+    fi
```
lpiwowar, line 326: ditto

lpiwowar (changes requested):
`tempet cleanup` command is failing on missing credentials:
```
+ tempest cleanup --init-saved-state
Failure during cleanup
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/tempest/cmd/cleanup.py", line 153, in take_action
    self.init(parsed_args)
  File "/usr/lib/python3.9/site-packages/tempest/cmd/cleanup.py", line 183, in init
    credentials.get_configured_admin_credentials())
  File "/usr/lib/python3.9/site-packages/tempest/common/credentials_factory.py", line 257, in get_configured_admin_credentials
    credentials = get_credentials(fill_in=fill_in,
  File "/usr/lib/python3.9/site-packages/tempest/common/credentials_factory.py", line 300, in get_credentials
    return auth.get_credentials(auth_url,
  File "/usr/lib/python3.9/site-packages/tempest/lib/auth.py", line 649, in get_credentials
    auth_provider = auth_provider_class(
  File "/usr/lib/python3.9/site-packages/tempest/lib/auth.py", line 263, in __init__
    super(KeystoneAuthProvider, self).__init__(credentials, scope)
  File "/usr/lib/python3.9/site-packages/tempest/lib/auth.py", line 91, in __init__
    raise exceptions.InvalidCredentials(message)
tempest.lib.exceptions.InvalidCredentials: Invalid Credentials
Details: Credentials are: {'domain_id': None, 'domain_name': 'Default', 'username': None, 'project_domain_id': None, 'project_domain_name': 'Default', 'project_id': None, 'project_name': None, 'tenant_id': None, 'tenant_name': None, 'user_domain_id': None, 'user_domain_name': 'Default', 'user_id': None, 'system': None} Password is not defined.
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/tempest/cmd/cleanup.py", line 153, in take_action
    self.init(parsed_args)
  File "/usr/lib/python3.9/site-packages/tempest/cmd/cleanup.py", line 183, in init
    credentials.get_configured_admin_credentials())
  File "/usr/lib/python3.9/site-packages/tempest/common/credentials_factory.py", line 257, in get_configured_admin_credentials
    credentials = get_credentials(fill_in=fill_in,
  File "/usr/lib/python3.9/site-packages/tempest/common/credentials_factory.py", line 300, in get_credentials
    return auth.get_credentials(auth_url,
  File "/usr/lib/python3.9/site-packages/tempest/lib/auth.py", line 649, in get_credentials
    auth_provider = auth_provider_class(
  File "/usr/lib/python3.9/site-packages/tempest/lib/auth.py", line 263, in __init__
    super(KeystoneAuthProvider, self).__init__(credentials, scope)
  File "/usr/lib/python3.9/site-packages/tempest/lib/auth.py", line 91, in __init__
    raise exceptions.InvalidCredentials(message)
tempest.lib.exceptions.InvalidCredentials: Invalid Credentials
```
and 
```
+ '[' false == false
/var/lib/tempest/run_tempest.sh: line 368: [: missing `]'
```

[1] https://logserver.rdoproject.org/08/208/f37dab74dbbd7d4cc910377ebf5f9d9a55965b6f/github-check/tcib-podified-multinode-edpm-deployment-crc/348e9cb/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/tempest-tests-xcclp/logs/tempest-tests-tests-runner.log

lpiwowar:
Missing brackets:
```
/var/lib/tempest/run_tempest.sh: line 332: syntax error in conditional expression
/var/lib/tempest/run_tempest.sh: line 332: syntax error near `]'
/var/lib/tempest/run_tempest.sh: line 332: `    if [[ ${TEMPEST_DEBUG_MODE} == false ] && [ ${TEMPEST_CLEANUP} == true ]]; then'
```

```
[[ ${TEMPEST_DEBUG_MODE} == false ]] && [[ ${TEMPEST_CLEANUP} == true ]]
```
(I missed the brackets in the comment above)

(Note I'm not sure whether this PR contains fix for the missing credentials error mentioned in the previous comment).

kopecmartin:
oh, the creds error is related to the fact that now we run 'tempest cleanup' before we have a tempest.conf .. we need to run discover-tempest-config first

lpiwowar:
> oh, the creds error is related to the fact that now we run 'tempest cleanup' before we have a tempest.conf .. we need to run discover-tempest-config first

We have to find either:
- how to run `tempest cleanup` first or
- how to ensure that the resources created by discover-tempest-config (especially cirros image) get cleaned up.

kopecmartin:
done, the easier way will be to generate a simple tempest.conf containing basically the default settings

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -318,10 +319,24 @@ function run_git_tempest {
     tempest init openshift
     pushd $TEMPEST_DIR
 
+    # We're running cleanup only under certain circumstances
+    if [[ ${TEMPEST_DEBUG_MODE} == false ]] && [[ ${TEMPEST_CLEANUP} == true ]]; then
+        # generate a simple tempest.conf so that we can run --init-saved-state
+        discover-tempest-config
```
lpiwowar, line 325: Won't this still keep the cirros image hanging?

```
(.venv) stack@lpiwowar:~/lpiwowar/tempest$ openstack image list

(.venv) stack@lpiwowar:~/lpiwowar/tempest$ discover-tempest-config
...
2024-09-18 12:17:39.957 3158161 INFO tempest.lib.common.rest_client [req-4f2727c6-d64e-4653-b029-28994322d87b req-4f2727c6-d64e-4653-b029-28994322d87b ] Request (main): 200 GET http://10.0.144.252/image/v2/images 0.011s
2024-09-18 12:17:39.957 3158161 INFO config_tempest.constants [-] Creating image 'cirros-0.6.2-x86_64-disk.img'
2024-09-18 12:17:39.957 3158161 INFO config_tempest.constants [-] Downloading 'https://download.cirros-cloud.net/0.6.2/cirros-0.6.2-x86_64-disk.img' and saving as '/opt/stack/lpiwowar/tempest/etc/cirros-0.6.2-x86_64-disk.img'
2024-09-18 12:17:59.038 3158161 INFO config_tempest.constants [-] Uploading image 'cirros-0.6.2-x86_64-disk.img' from '/opt/stack/lpiwowar/tempest/etc/cirros-0.6.2-x86_64-disk.img'
2024-09-18 12:17:59.274 3158161 INFO tempest.lib.common.rest_client [req-8a28c07e-8e5c-4d07-afb7-ba813f53effb req-8a28c07e-8e5c-4d07-afb7-ba813f53effb ] Request (main): 201 POST http://10.0.144.252/image/v2/images 0.235s
2024-09-18 12:17:59.848 3158161 INFO tempest.lib.common.rest_client [req-87ff9abf-8d43-4b24-bd70-76da9f1d13cd req-87ff9abf-8d43-4b24-bd70-76da9f1d13cd ] Request (main): 204 PUT http://10.0.144.252/image/v2/images/0b10b998-e455-4829-90ae-e70aef8cc6e0/file 0.574s
...
(.venv) stack@lpiwowar:~/lpiwowar/tempest$ openstack image list
+--------------------------------------+----------------------------------+--------+
| ID                                   | Name                             | Status |
+--------------------------------------+----------------------------------+--------+
| 0b10b998-e455-4829-90ae-e70aef8cc6e0 | cirros-0.6.2-x86_64-disk.img     | active |
| 4cbfd5fe-f715-4a83-b505-19262e1115e1 | cirros-0.6.2-x86_64-disk.img_alt | active |
+--------------------------------------+----------------------------------+--------+
```

lpiwowar (commented):
One question.

container-images/tcib/base/os/tempest/run_tempest.sh
kopecmartin: yup, we have 2 options then:
1. make discover-tempest-config not to create any images, flavors etc, just a very simple tempest.conf
2. execute 'openstack image delete cirros-0.6.2-x86_64-disk.img' and 'openstack image delete cirros-0.6.2-x86_64-disk.img_alt' commands

I'll check the options and get back to you

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -328,10 +329,29 @@ function run_git_tempest {
...
-    discover_tempest_config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
+    # We're running cleanup only under certain circumstances
+    if [[ ${TEMPEST_DEBUG_MODE} == false ]] && [[ ${TEMPEST_CLEANUP} == true ]]; then
+        # generate a simple tempest.conf so that we can run --init-saved-state
+        discover-tempest-config
+        # let's remove the images that discover-tempest-config creates by default
+        # so that the're not part of the saved_state.json and can be deleted
+        # by tempest cleanup later
+        openstack image delete cirros-0.6.2-x86_64-disk.img
+        openstack image delete cirros-0.6.2-x86_64-disk.img_alt
```
ashu-011, line 340: If we are indeed going by this approach maybe you can just do something like 
`openstack image list -c Name -f value | xargs -I {} openstack image delete {}` 
That way we won't be hardcoding values  in case image name changes or we use a newer version in future  .

lpiwowar: This sounds like a good idea! My only concern is that it deletes all images. Maybe we can update the command with grep for cirros?

```
openstack image list -c Name -f value 
   | grep cirros 
   | xargs -I {} openstack image delete {}
```

container-images/tcib/base/os/tempest/run_tempest.sh
lpiwowar: Ok, got it. I agree it seems using the openstack CLI is the best solution right now.

container-images/tcib/base/os/tempest/run_tempest.sh
kopecmartin: sounds good, let's do that, it's better as we won't need to update it with every cirros version update in tempestconf

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -328,10 +329,28 @@ function run_git_tempest {
     tempest init openshift
     pushd $TEMPEST_DIR
 
-    discover_tempest_config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
+    # We're running cleanup only under certain circumstances
+    if [[ ${TEMPEST_DEBUG_MODE} == false ]] && [[ ${TEMPEST_CLEANUP} == true ]]; then
```
lpiwowar, line 333: @kopecmartin, do we really want the cleanup to be cancelled if `debug` mode is turned on? What if somebody wants to run in debug mode and keep `cleanup` turned on. 

If we do this we can leave the decision up to the user.
```
if [[ ${TEMPEST_CLEANUP} == true ]]; then
```

lpiwowar (commented):
Looks good to me! :+1: 

I've just added one comment. What do you think @kopecmartin?

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -328,10 +329,28 @@ function run_git_tempest {
...
+        # let's remove the images that discover-tempest-config creates by default
+        # so that the're not part of the saved_state.json and can be deleted
+        # by tempest cleanup later
+        openstack image list -c Name -f value | grep cirros | xargs -I {} openstack image delete {}
+        tempest cleanup --init-saved-state
+    fi
+
+    upload_extra_images
+
+    discover-tempest-config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
```
lpiwowar, line 345:
```suggestion
    discover_tempest_config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
```

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -350,10 +369,28 @@ function run_rpm_tempest {
...
```
lpiwowar, line 385:
```suggestion
    discover_tempest_config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
```

container-images/tcib/base/os/tempest/run_tempest.sh
kopecmartin: thanks! i totally missed the - vs _ change

lpiwowar (approved):
Looks good to me! :+1:

ashu-011 (approved):

PR #209, opened 2024-09-09
Avoid malformed results xml

kopecmartin:
In case there are no results, the xml files get malformed. The malformed xml is then refused by polarion.

This edits the code the way the xml and html results will be generated only in the case when the subunit was generated with 0 return code (successfull operation).

https://issues.redhat.com/browse/OSPRH-9242

### DIFF
--- a/container-images/tcib/base/os/tempest/run_tempest.sh
+++ b/container-images/tcib/base/os/tempest/run_tempest.sh
@@ -364,14 +364,10 @@ function generate_test_results {
     TEMPEST_LOGS_DIR=${TEMPEST_PATH}${TEMPEST_WORKFLOW_STEP_DIR_NAME}/
     mkdir -p ${TEMPEST_LOGS_DIR}
 
-    echo "Generate subunit"
-    stestr last --subunit > ${TEMPEST_LOGS_DIR}testrepository.subunit || true
-
-    echo "Generate subunit xml file"
-    subunit2junitxml ${TEMPEST_LOGS_DIR}testrepository.subunit > ${TEMPEST_LOGS_DIR}tempest_results.xml || true
-
-    echo "Generate html result"
-    subunit2html ${TEMPEST_LOGS_DIR}testrepository.subunit ${TEMPEST_LOGS_DIR}stestr_results.html || true
+    echo "Generate subunit, then xml and html results"
+    stestr last --subunit > ${TEMPEST_LOGS_DIR}testrepository.subunit \
+    && (subunit2junitxml ${TEMPEST_LOGS_DIR}testrepository.subunit > ${TEMPEST_LOGS_DIR}tempest_results.xml || true) \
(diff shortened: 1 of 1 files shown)


container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -364,14 +364,10 @@ function generate_test_results {
...
 
-    echo "Generate subunit"
-    stestr last --subunit > ${TEMPEST_LOGS_DIR}testrepository.subunit || true
-
-    echo "Generate subunit xml file"
-    subunit2junitxml ${TEMPEST_LOGS_DIR}testrepository.subunit > ${TEMPEST_LOGS_DIR}tempest_results.xml || true
-
-    echo "Generate html result"
-    subunit2html ${TEMPEST_LOGS_DIR}testrepository.subunit ${TEMPEST_LOGS_DIR}stestr_results.html || true
+    echo "Generate subunit, then xml and html results"
```
kopecmartin, line 367: validation of the && and || logic, all possible situations:

```
# zero return code and successfully executes both of the echos - ideal case
$ true && (echo 1 || true) && echo 2 || true
1
2

# zero return code and no echo outputs - that's what we need to fix the issue 
$ false && (echo 1 || true) && echo 2 || true

# zero return code and only the second echo successful - the first echo didn't affect the second one
$ true && (false || true) && echo 2 || true
2

# zero return code and only the first echo successful - the second echo didn't affect the first one nor the overall result
$ true && (echo 1 || true) && false || true
1
$ 
```

lpiwowar: What do you think about this version?

```
stestr last --subunit > ${TEMPEST_LOGS_DIR}testrepository.subunit

if [[ $? -eq  0 ]]; then
  subunit2junitxml ${TEMPEST_LOGS_DIR}testrepository.subunit > ${TEMPEST_LOGS_DIR}tempest_results.xml
  subunit2html ${TEMPEST_LOGS_DIR}testrepository.subunit ${TEMPEST_LOGS_DIR}stestr_results.html
fi
```

lpiwowar (commented):
One small comment.

container-images/tcib/base/os/tempest/run_tempest.sh
kopecmartin: same result with the difference i need to modify the patch one more time :D

kopecmartin:
what are we waiting for here? it's ready to be merged or I need to hear specific feedback

container-images/tcib/base/os/tempest/run_tempest.sh
lpiwowar: Yes, the result is the same. Readability wise, I'm not so sure. Take a look on how long comment you had to make just to explain how it works.

I'm not holding this. Let's move this forward.

lpiwowar (approved):
LGTM :+1:

PR #210, opened 2024-09-10
[tempest] Handle quotes in TEMPESTCONF_OVERRIDES

ASBishop:
Use 'eval' to expand the TEMPESTCONF_OVERRIDES variable prior to using it to run discover-tempest-config. This prevents embedded quotes within the variable from being mangled.

Without the 'eval' embedded quoted strings like "Some String" get mangled into '"Some' 'String"' on the resulting command line.

Jira: [OSPRH-10092](https://issues.redhat.com//browse/OSPRH-10092)


kopecmartin (approved):
lgtm

PR #212, opened 2024-09-18
[tempest] Fix special chars in TEMPESTCONF_OVERRIDES

ASBishop:
Fix a regression introduced in PR #210 in which strings with special bash characters need to be escaped: cubswin:) has to be `cubswin:\)`. The need to escape some characters breaks existing CI jobs.

This patch fixes the regression so it's not necessary to escape special characters.

### DIFF
--- a/container-images/tcib/base/os/tempest/run_tempest.sh
+++ b/container-images/tcib/base/os/tempest/run_tempest.sh
@@ -288,6 +288,16 @@ function upload_extra_images {
     done
 }
 
+
+# This function ensures all arguments are handled properly:
+# - Embedded quotes are preserved, e.g. "Some string"
+# - Special bash characters don't need to be escaped, e.g. cubswin:)
+function discover_tempest_config {
+    cat <<EOF | xargs discover-tempest-config
+$*
+EOF
+}
+
 function run_git_tempest {
     mkdir -p $TEMPEST_EXTERNAL_PLUGIN_DIR
     pushd $TEMPEST_EXTERNAL_PLUGIN_DIR
@@ -318,7 +328,7 @@ function run_git_tempest {
     tempest init openshift
     pushd $TEMPEST_DIR
 
-    eval discover-tempest-config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
+    discover_tempest_config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
     && tempest run ${TEMPEST_ARGS}
     RETURN_VALUE=$?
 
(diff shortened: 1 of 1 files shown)


container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -288,6 +288,16 @@ function upload_extra_images {
...
+
+# This function ensures all arguments are handled properly:
+# - Embedded quotes are preserved, e.g. "Some string"
+# - Special bash characters don't need to be escaped, e.g. cubswin:)
+function discover_tempest_config {
+    cat <<EOF | xargs discover-tempest-config
+$*
+EOF
+}
+
```
SeanMooney, line 300: this seam to work as indended 
```
00:41:58]➜ function discover_tempest_config {
    cat <<EOF | xargs discover-tempest-config
$*
EOF
 }
~ via 🐹 v1.20.12 via 🐍 v3.11.9 
[00:42:04]➜ discover_tempest_config test
test
~ via 🐹 v1.20.12 via 🐍 v3.11.9 
[00:42:07]➜ discover_tempest_config test:)
bash: syntax error near unexpected token `)'
~ via 🐹 v1.20.12 via 🐍 v3.11.9 
[00:42:10]❯ data="test:)"
~ via 🐹 v1.20.12 via 🐍 v3.11.9 
[00:42:30]➜ discover_tempest_config ${data}
test:)
~ via 🐹 v1.20.12 via 🐍 v3.11.9 
[00:42:39]➜ data="test :)"
~ via 🐹 v1.20.12 via 🐍 v3.11.9 
[00:42:57]➜ discover_tempest_config ${data}
test :)
~ via 🐹 v1.20.12 via 🐍 v3.11.9
```
at least if the charters that need to be escapes are in a variable
or they are quoted when being passed to the function directly.

```
[00:42:58]➜ discover_tempest_config '${data}'
${data}
~ via 🐹 v1.20.12 via 🐍 v3.11.9
[00:42:58]➜ discover_tempest_config '${data}'
${data}
~ via 🐹 v1.20.12 via 🐍 v3.11.9 
```
interpolation or not of variable in string also works

```
[00:45:31]❯ data="test; echo do bad thing :("
~ via 🐹 v1.20.12 via 🐍 v3.11.9 
00:47:04]➜ discover_tempest_config "${data}"
test; echo do bad thing :(
~ via 🐹 v1.20.12 via 🐍 v3.11.9 
[00:47:08]➜ discover_tempest_config ${data}
test; echo do bad thing :(
```
it also prevent basic exploring like adding `; shell injection goes here`

not that it nessicaly safe but i think this is safer then eval too.

gibizer (approved):

mrkisaolamb (approved):
lgtm

lpiwowar (approved):
LGTM! :+1:

SeanMooney (approved):

bshephar (approved):

rabi (approved):
/lgtm

PR #221, opened 2024-10-21
Create 2 dummy temporary flavors

kopecmartin:
discover-tempest-config needs 2 flavors it can't run without. When ran without "--create" param, it can't create the flavors itself and fails.

Let's create 2 flavors and delete them afterwards to leave the system intact.

### DIFF
--- a/container-images/tcib/base/os/tempest/run_tempest.sh
+++ b/container-images/tcib/base/os/tempest/run_tempest.sh
@@ -331,8 +331,14 @@ function run_git_tempest {
 
     # We're running cleanup only under certain circumstances
     if [[ ${TEMPEST_CLEANUP} == true ]]; then
+        # discover-tempest-config needs 2 flavors it can't run without. When ran without "--create"
+        # param, it can't create the flavors itself and fails.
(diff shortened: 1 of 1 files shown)


lpiwowar:
Testing here -> https://github.com/openstack-k8s-operators/test-operator/pull/216

lpiwowar (approved):
Looks good to me! :+1: 

I've tested the change here and it seems to be working [1]. It would be nice if there was other solution than creating and deleting the flavors but we've already discussed this with Martin and this is probably the best solution right now. 

There is a failure in deletion of the `Quota for tenant` [1] which is not related to this PR and can be addressed in a future if it turns out to be a problem.

[1] [here](https://logserver.rdoproject.org/16/216/41e791119dfe5774b007e5849834717f9fdcbd1a/github-check/podified-multinode-edpm-deployment-crc-test-operator/e2c034b/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/tempest-tests-full-workflow-step-0-qxr9n/logs/tempest-tests-tests-runner.log)

lpiwowar:
recheck

PR #228, opened 2024-11-12
Enable the netutils iamge

beagles:
Patch enables building the image definition previously introduced in PR 218.


beagles:
Basically forgot the add the containers.yaml to the previous commit.

abays (approved):
/lgtm

PR #229, opened 2024-11-15
Add inotify-tools to ironic-inspector image

steveb:
Both ironic-conductor and ironic-inspector have sidecars which cat the ramdisk logs, and these are being converted to use inotifywait.

The conductor image already has the inotify-tools package, this commit adds it to the ironic-inspector image too.

Jira: [OSPRH-10240](https://issues.redhat.com//browse/OSPRH-10240)


rabi (approved):
/lgtm

PR #250, opened 2025-02-03
Add sudoers file in ansible-test container

eshulman2:
Ansible-test container was missing the sudoers file in it, added a copy statment in the ansibleTest.yaml to add it to the image.


eshulman2:
Tested with locally build container

PR #253, opened 2025-02-17
[DNM] try build multi stage build with tempest promtool

raukadah:


arxcruz:
NO, please just NO, this is a VERY dirty workaround, and probably also security breach. Just create a new tempest-prometheus if you need it.

PR #261, opened 2025-03-12
Fix expectedFailuresList incorrect overwrite

kstrenkova:
During a job run, we have noticed that the job status was reported as success despite an error occurring. The root cause was that stestr faling list was empty due to an error arising before any tests were executed. However, expectedFailuresList mistakenly interpreted this as if there were no failing tests. This patch aims to fix this edge case.


container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -482,7 +482,7 @@ if [ ${TEMPEST_DEBUG_MODE} == true ]; then
     sleep infinity
 fi
 
-if [ ! -z ${TEMPEST_EXPECTED_FAILURES_LIST} ]; then
+if [ -s ${TEMPEST_LOGS_DIR}stestr_failing.txt ] && [ ! -z ${TEMPEST_EXPECTED_FAILURES_LIST} ]; then
```
sdatko, line 485: I am good with the first condition here, i.e. the one that was added, but I think the second one should be corrected – from what we saw, the `${TEMPEST_EXPECTED_FAILURES_LIST}` is a path to file with tests that we accept to fail; `-z` checks for string length to be zero, hence, this condition will be always true currently I think? Shouldnt it also be `-s` simply?

From man page for `test`:
```
-z STRING
       the length of STRING is zero

-s FILE
       FILE exists and has a size greater than zero
```

kstrenkova: Thanks for the review! I had to go back and look how the expectedFailuresList is created. Yes, it does create file expected_failures.txt so checking the size of file instead of the string could be done. I will change it locally and try some simple test if it indeed works.

Post thoughts: I find it interesting that includeList and excludeList are done the same way (with -z) :thinking: I was just wondering why it is used in such way because I took them as reference when creating the code.

kstrenkova: But also thinking about it more, I want to propose a change that will change expectedFailuresList default('') to default(omit) :thinking: Which would mean the -z check would be relevant again and we would keep the same pattern between all three lists which are not lists :D (include, exclude, expectedFailures). What do you think?

sdatko: Because Bash is full of surprises and many people do not understand how it works; then simply as you said, often repeat what already exist, without checking if that makes sense :wink: Or maybe simply initially in PoC it was passed as a string (see the variable name do not suggest to be a file), but then when the argument became too large in production usage so it landed in a file, yet the check was never corrected. Nevertheless, we caught that, so we can improve :heart:

kstrenkova: Ack, I changed it to checking the file and its size :+1: Regarding the impact of the default(omit) I am not sure myself. We can discuss this on the patch itself again :D

sdatko (approved):
Looks good to me; let's wait for the final confirmation from tests before merging, but as to my intuition it should work properly now :+1:

kstrenkova:
I am back with finished local testing. I have tested these scenarios:

- No test failed. We correctly get return value 0.
- Tests have failed and all are in expectedFailuresList. We correctly get return value 0.
- Test have failed, but at least one was not in expectedFailuresList. We correctly get non-zero return value.
- There is an error before running any tests. We correctly get non-zero return value.

TL;DR The change is running correctly :partying_face:

kstrenkova:
recheck

sdatko:
recheck :thinking:

sdatko:
recheck

abays (approved):
/lgtm

PR #263, opened 2025-03-18
Allow installing not only collection

eshulman2:
when using collection inside the ansible-galaxy command with the collection flag only collections are installed this doesn't install additional types of possible requirments like roles. removed the collection keyword from the ansible galaxy install command to allow installing roles as requirements

### DIFF
--- a/container-images/tcib/base/ansible-tests/run_ansible.sh
+++ b/container-images/tcib/base/ansible-tests/run_ansible.sh
@@ -35,7 +35,7 @@ fi
 
 # Install collections from requirements.yaml if the file exists
 if [[ -f "$ANSIBLE_DIR/requirements.yaml" ]]; then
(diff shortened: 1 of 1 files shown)


rabi:
collections are a way to package and distribute reusable content like roles, modules, and plugins. I'm not sure why you would want to install roles directly from requirements.yaml. But this change does not harm.

rabi (approved):
/lgtm

eshulman2:
@rabi In case you would like to import a specific role from a git project without having to upload it to galaxy as a collection for example

rabi:
> @rabi this is useful in case where you want to import a specific role from a git project without having to upload it to galaxy as a collection for example

Generally you should not be doing that. All standard roles are published to galaxy as collections. This is a test container so it's fine.

PR #269, opened 2025-04-29
Add CentOS Stream 10 support in tcib

raukadah:
This pr updates the container files for centos stream 10:

-  In CS10, python3.12 is the default python runtime that's why tcib_python_version is used in horizon
-  mod_auth_mellon is dropped from CS10, that's why it is default to cs9 only.
-    tobiko container needs libffi-devel package to be there.

Tested here: https://softwarefactory-project.io/zuul/t/rdoproject.org/build/3137cbfb56cd4445ad1f2805cd1c4e96 and here is the built containers: https://logserver.rdoproject.org/313/rdoproject.org/3137cbfb56cd4445ad1f2805cd1c4e96/ci-framework-data/logs/containers-built.log

Jira: [OSPRH-16308](https://issues.redhat.com//browse/OSPRH-16308)


raukadah:
recheck

deshipu (commented):
I can't see anything obviously wrong there.

bshephar (approved):
Seems reasonable to me

PR #270, opened 2025-04-29
Set git config http.sslVerify=false

HanzJas:
Currently there is no possibility to clone Horizon repository from GitLab because of SSL certificate problem.


ashu-011 (approved):

abays (approved):
Workaround seems okay for now

/lgtm

PR #274, opened 2025-05-15
Update urllib version for horizontest container

ashu-011:
Update urllib3 version for it be compatible with Centos10 . With current version there is a conflict 
Resolves https://issues.redhat.com/browse/OSPRH-16481


raukadah:
recheck

PR #282, opened 2025-06-05
Replace mod_wsgi with uwsgi for designateapi

omersch381:
mod_wsgi is being removed from upstream sources but OpenStack APIs are still using it downstream, which is very concerning.

This patch replaces mod_wsgi with uwsgi.

### DIFF
--- a/container-images/tcib/base/os/designate-base/designate-api/designate-api.yaml
+++ b/container-images/tcib/base/os/designate-base/designate-api/designate-api.yaml
@@ -6,4 +6,5 @@ tcib_packages:
   - httpd
   - mod_ssl
   - python3-mod_wsgi
+  - uwsgi
 tcib_user: designate


bogdando (commented):
multiple services do use mod_wsgi. If there is upgrade impact for rhel/centos 10 related to this package, please create a jira issue and link it here.
My concern is that we should get this acked by the release delivery team, I am not sure uwsgi switch would be straightforward

@SeanMooney  FYI

SeanMooney:
> mod_wsgi is being removed from upstream sources but OpenStack APIs are still using it downstream, which is very concerning.
> 
> This patch replaces mod_wsgi with uwsgi.

that is not actully correct.

we simply changed how we package the wsgi applcation with in the git trees.
this does not alter whether mod_wsgi is supprot or not.


for what it worth i would like to see us move to gunicorn, uvicorn or uwsgi.

so i dont nessiarly see a problem with designate moving now.

we filed https://issues.redhat.com/browse/RHOSSTRAT-220 in 2023 becasue we wanted to intially od this before 18 released to slim down our contianers.

for nova/placement the switch woudl be pretty easy provide we have the package.

the main concern is with ssl termination and ensuring we can still terminate it in the api pod.
terminitating ssl at the openshfit router only instead of in the api pod would be a regersssion form a secuirty point of view and runnign appache just for tls termination in a side care contaienr to delegate to uwsgi like we do in devstack woud defeat orginal goal of slimig down the continers.

designate-operator can simply provide wa wsgi file or point mod_wsgi at the python module that was added upstream as part fo addign supprot for pyproject.toml and the module based wsgi supprot

cmsj:
@SeanMooney I think that Jira probably needs to be swapped out for something else - it's asking for action before we even know what that action would be and whether release teams can support it. I don't see any pressing need to do this in 18 though?

container-images/tcib/base/os/designate-base/designate-api/designate-api.yaml
```diff
@@ -5,5 +5,5 @@ tcib_packages:
   - openstack-designate-api
   - httpd
   - mod_ssl
-  - python3-mod_wsgi
```
fmount, line 8: maybe we can keep `python3-mod_wsgi`, which is installed by other services as well, and keep backward compatibily, which means we can choose the deployment strategy from the operator (proxypass vs mod_wsgi).

omersch381: I agree, thanks!

PR #283, opened 2025-06-06
Add pre-ansible-playbook hook to run arbitrary commands

lmiccini:


lmiccini:
will revisit together with the respective test-operator change.

PR #284, opened 2025-06-06
[ANTELOPE] Added CloudKitty to tcib to produce api and processor images

jlarriba:


container-images/tcib/base/os/horizon/horizon.yaml
```diff
@@ -34,3 +34,4 @@ tcib_packages:
   - openstack-octavia-ui
   - openstack-designate-ui
   - openstack-watcher-ui
+  - openstack-cloudkitty-ui
```
raukadah, line 37: For watcher-dashboard, here is what we added https://github.com/openstack-k8s-operators/tcib/commit/bd2fa10fc5bdb24eca21a520ee24870502effa84.

I need we also similar approach for cloudkitty-ui. 
Reason I am posting this comment, If horizon gets deployed and cloudkitty-ui package is installed then cloudkitty panel will appear in the Horizon dashboard but cloudkitty service is not configured.

stuggi (approved):
/lgtm

PR #287, opened 2025-06-19
Exclude federation tests by default

ashu-011:
This commit is excluding federation tests by default as part of extra param because they are only meant to run in the job enabling federation/OIDC support.


HanzJas (approved):
LGTM, thank you!

abays (approved):
/lgtm

PR #294, opened 2025-07-03
Remove TOBIKO_UPDATE_REPO logic; always update Tobiko repo

belolipa:
Always updating the Tobiko repository is the preferred behavior. Removing the TOBIKO_UPDATE_REPO variable entirely would eliminate unnecessary complexity.


eduolivares:
recheck

lpiwowar (approved):

PR #296, opened 2025-07-08
Change method of logging for test-operator

kstrenkova:
This patch introduces a new way of logging test-operator tests. Before, the logs were generated after the test run, which resulted in limited log information when the test execution timed out. Now, the logs are gathered continuously.


rabi (approved):
/lgtm

PR #297, opened 2025-07-15
[tobiko] install required packages after removing python3-chardet

eduolivares:
The tobiko image needs to remove python3-chardet because of problems
with dependencies installing tox later. However, other required
packages may be removed too during the removal of python3-chardet.

The solution is to simply reorder the installation steps, guaranteeing
that the required packages are not removed.


sdatko:
recheck

sdatko (approved):

PR #303, opened 2025-07-23
[tobiko] Downgrade python-dateutil to make it work with cs9/10

raukadah:
python-dateutil-2.9.0 is present in cs10 but python-dateutil-2.8 version is present in both cs9 and cs10.

if we remove python-dateutil in cs10, it also removes python3-dnf-plugins-core package which might break other stuff.

By downgrading python-dateutil package, it works for cs10.

Signed-off-by: Chandan Kumar (raukadah) <raukadah@gmail.com>

Resolves: OSPCIX-987


container-images/tcib/base/tobiko/tobiko.yaml
```diff
@@ -8,6 +8,8 @@ tcib_actions:
 - run: >-
     if [ '{{ tcib_distro }}' == 'rhel' ];then
     if [ -n "$(rpm -qa redhat-release)" ];then dnf -y remove python3-chardet; fi ; fi
+- run: >-
+    if [ '{{ tcib_release }}' == '10' ];then dnf -y downgrade python-dateutil; fi
```
rlandy, line 12: If this is a temp solution, let's add a comment to say we have to revert

PR #316, opened 2025-08-14
Add reusing Tempest stestr timing data

kstrenkova:
This patch aims to help jobs save time when executing Tempest
tests. This is done by reusing stestr timing data that are saved
during the run.

### DIFF
--- a/container-images/tcib/base/os/tempest/run_tempest.sh
+++ b/container-images/tcib/base/os/tempest/run_tempest.sh
@@ -349,6 +349,10 @@ function run_tempest {
 
     upload_extra_images
 
+    if [[ -n "${TEMPEST_TIMING_DATA_URL:-}" ]]; then
+        curl -L "${TEMPEST_TIMING_DATA_URL}" | tar -xz --strip-components=1 -C "${TEMPEST_DIR}/.stestr"
+    fi
+
     mkdir -p "${TEMPEST_LOGS_DIR}"
 
     discover_tempest_config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
@@ -444,7 +448,7 @@ function save_config_files {
     cp -f "${TEMPEST_DIR}/etc/"*.{conf,ini,txt,yaml} "${TEMPEST_LOGS_DIR}/etc"
 
     cp -f "${TEMPEST_DIR}/.stestr.conf" "${TEMPEST_LOGS_DIR}/stestr.conf"
-    cp -rf "${TEMPEST_DIR}/.stestr" "${TEMPEST_LOGS_DIR}/stestr"
+    tar -czf "${TEMPEST_LOGS_DIR}/stestr.tar.gz" -C "${TEMPEST_DIR}" .stestr
 }
 
 


container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -457,6 +470,21 @@ function generate_test_results {
...
+    mkdir -p ${FULL_PATH}
+    cp -r ${TEMPEST_STESTR}/* ${FULL_PATH}
+
+    git config --global user.email "dev@null.io"
+    git config --global user.name "Dummy"
+
+    cd ${HOMEDIR}/stestr-repo
+    git add ${STESTR_REPO_DIR}
+    git commit -m "Test update ${STESTR_REPO_DIR}."
+    git push origin main
```
sdatko, line 485: We may need to implement some sort of conflict-resolution/rebase&retry e.g. in case of multiple instances of the same job running at close/same time (or I am thinking too much?)

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -457,6 +470,21 @@ function generate_test_results {
...
 }
 
+function push_timing_data {
+    git clone --branch main ${STESTR_REPO_URL} ${HOMEDIR}/stestr-repo
+    FULL_PATH=${HOMEDIR}/stestr-repo/${STESTR_REPO_DIR}
+    mkdir -p ${FULL_PATH}
+    cp -r ${TEMPEST_STESTR}/* ${FULL_PATH}
+
+    git config --global user.email "dev@null.io"
+    git config --global user.name "Dummy"
```
sdatko, line 480: TODO: make that configurable?

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -291,6 +295,11 @@ function upload_extra_images {
     done
 }
 
+function pull_timing_data {
```
sdatko, line 298: TODO: make this mechanism optional (here and in `push_timing_data` ~ skip by default?)

I imagine we may want to ignore some runs, e.g. from testprojects

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -97,6 +97,10 @@ TEMPEST_ARGS=""
 TEMPEST_DEBUG_MODE="${TEMPEST_DEBUG_MODE:-false}"
 TEMPEST_CLEANUP="${TEMPEST_CLEANUP:-false}"
 
+TEMPEST_STESTR="$HOMEDIR/openshift/.stestr"
+STESTR_REPO_URL="https://gitlab.cee.redhat.com/kstrenko/tempest-stestr-data.git"
+STESTR_REPO_DIR="components-integration/component-cinder-edpm-rhel9-rhoso18.0-crc-nfs/tempest-tests-tempest"
```
sdatko, line 102: TODO: Those parameters will be set via Zuul job and passed here, by default we want to keep just `example.com/repo.git` and `tenant/job/framework` here I guess

container-images/tcib/base/os/tempest/run_tempest.sh
kstrenkova: For pulling the data I have a parameter called TimingData ready that would serve as a boolean flag. For the data pushing I am not sure. Should I create a separate parameter that would be specifically for pushing?

container-images/tcib/base/os/tempest/run_tempest.sh
kstrenkova: For sure. I just encountered this error locally and wasn't sure if I want to overcomplicate the solution for test-run only. Now that more things we need are accumulating, maybe a struct for `TimingData` would make more sense. Like `TimingData: { Pull, Push, Username, UserEmail ,...}`.

container-images/tcib/base/os/tempest/run_tempest.sh
kstrenkova: I didn't think of that :thinking: If multiple runs of the same job can be at the same time, maybe we could turn off the `git push` for all of them except one? I am imagining it that they are being called from different places, but if that's not the case then a retry could be needed.

container-images/tcib/base/os/tempest/run_tempest.sh
kstrenkova: Yes, of course. For now it is hard-coded for testing purposes. If we see that the timing data don't help much, creating multiple new variables would be work for naught.

kstrenkova (commented):
Thanks for the review @sdatko! This is a very [DNM] [Testing] change, so I agree with all of yours todos. I am really curious if even this version will work and tell us how much stestr timing data could help us with reducing the execution time. Hopefully it will!

container-images/tcib/base/os/tempest/run_tempest.sh
sdatko: Ok, I was overthinking probably. Probably we just want to keep the timing data from the last successful run of a job in promotions pipeline, not testprojects etc. Especially when looking at the data file format we checked recently during the call. We will not have to bother rebasing etc then.

But as per https://github.com/openstack-k8s-operators/tcib/pull/316/files#r2293669982 – let's move those considerations somewhere else and make run_tempest simple.

container-images/tcib/base/os/tempest/run_tempest.sh
sdatko: My main point of this comment was to make a remainder ~ if the timing data are not available under given URL or were not passed, this script should not fail and just proceed with clean run (with no timing data) as it was so far.

About the push – let's just remove it from this script at all, see: https://github.com/openstack-k8s-operators/tcib/pull/316/files#r2293669982

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -349,6 +349,10 @@ function run_tempest {
 
     upload_extra_images
 
+    if [[ "${TEMPEST_TIMING_DATA_URL}" ]]; then
+        curl -L "${TEMPEST_TIMING_DATA_URL}" | tar -xz -C "${TEMPEST_DIR}/.stestr"
```
sdatko, line 353: this `tar` call will fail if `.stestr` do not exist ~ please just use `-C "${TEMPEST_DIR}"` and see my comment below about archiving the whole directory

sdatko (changes requested):

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -349,6 +349,10 @@ function run_tempest {
 
     upload_extra_images
 
+    if [[ "${TEMPEST_TIMING_DATA_URL}" ]]; then
```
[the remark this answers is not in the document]
kstrenkova: Good catch. Didn't notice I didn't do `-n`. I added the empty default as well.

container-images/tcib/base/os/tempest/run_tempest.sh
kstrenkova: I went with using `--strip-components=1`. The `.stestr` directory is created during `tempest init openshift`, so it should always be present in that moment in code.

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -349,6 +349,10 @@ function run_tempest {
 
     upload_extra_images
 
+    if [[ -n "${TEMPEST_TIMING_DATA_URL:-}" ]]; then
+        curl -L "${TEMPEST_TIMING_DATA_URL}" | tar -xz --strip-components=1 -C "${TEMPEST_DIR}/.stestr"
```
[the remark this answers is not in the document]
kstrenkova: Thanks for the review! Personally, I would leave it as it is, even though the `mkdir` could act as a safety net.

rabi (approved):
/lgm

hjensas (approved):
Nice one!

PR #317, opened 2025-08-14
Add tobikoPatch parameter for custom patches

belolipa:
This pull request introduces a new parameter, tobikoPatch, to the Tobiko Custom Resource to allow for testing against specific. [OSPRH-11286]


container-images/tcib/base/tobiko/run_tobiko.sh
```diff
@@ -24,6 +24,8 @@ TOBIKO_VERSION=${TOBIKO_VERSION:-master}
 TOBIKO_PRIVATE_KEY_FILE=${TOBIKO_PRIVATE_KEY_FILE:-id_ecdsa}
 TOBIKO_KEYS_FOLDER=${TOBIKO_KEYS_FOLDER:-${TOBIKO_DIR}/external_files}
 TOBIKO_LOGS_DIR_NAME=${TOBIKO_LOGS_DIR_NAME:-"tobiko"}
+TOBIKO_PATCH_REPOSITORY="${TOBIKO_PATCH_REPOSITORY:-}"
```
kstrenkova, line 27: These are not setting any default, so I don't think they are needed.

container-images/tcib/base/tobiko/run_tobiko.sh
```diff
@@ -41,6 +43,10 @@ pushd tobiko
 git pull --rebase
 git checkout ${TOBIKO_VERSION}
 
+if [[ -n "$TOBIKO_PATCH_REFSPEC" ]]; then
```
kstrenkova, line 46: From the comment @eduolivares left on the ticket, I would say this `if condition` should decide if we use the new approach or if we use `git checkout ${TOBIKO_VERSION}`. So this is not the right place imho.

Additionally, to keep consistency in the code, I would suggest using this formatting: `${VARIABLE}` instead of this formatting: `"$VARIABLE"`.

kstrenkova (changes requested):
Nice first patch! There are some minor details to be fixed, but after fixing them I think it should work \o/ We will know after some local testing.

container-images/tcib/base/tobiko/run_tobiko.sh
```diff
@@ -39,7 +39,11 @@ cp -r ${TOBIKO_SRC_DIR} tobiko
 chown tobiko:tobiko -R tobiko
 pushd tobiko
 git pull --rebase
-git checkout ${TOBIKO_VERSION}
+if [ -n "${TOBIKO_PATCH_REFSPEC}" ]; then
+    git fetch ${TOBIKO_PATCH_REPOSITORY} ${TOBIKO_PATCH_REFSPEC} && git checkout FETCH_HEAD
+else
+    git checkout ${TOBIKO_VERSION}
```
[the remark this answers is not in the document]
sdatko: Ack, fine for me, though I am in favor of just introducing already better code immediately (especially that here is it a little inconsistent), but ok, let's do in separate change (I admit I also didnt connect in my mind that it is in Tobiko, not Tempest where I was fixing that recently :sweat:)

sdatko (approved):

adrianfusco (approved):

PR #318, opened 2025-08-19
Fix issue TEMPEST_EXPECTED_FAILURES_LIST is empty

hjensas:
Because the condition [ -s ${var} ] would evaluate to [ -s  ], with an empty argument, the shell can behave unexpectedly, and subsequent grep commands would try to read from an empty filename.

We belive this is the reason failing tempest tests did not fail, see Jira for logs.

Add a check to ensure TEMPEST_EXPECTED_FAILURES_LIST is not empty before attempting file operations.

Also add quoting around variable references.

Fixes: OSPRH-19179

### DIFF
--- a/container-images/tcib/base/os/tempest/run_tempest.sh
+++ b/container-images/tcib/base/os/tempest/run_tempest.sh
@@ -499,10 +499,10 @@ if [ ${TEMPEST_DEBUG_MODE} == true ]; then
     sleep infinity
 fi
 
-if [ -s ${TEMPEST_LOGS_DIR}stestr_failing.txt ] && [ -s ${TEMPEST_EXPECTED_FAILURES_LIST} ]; then
+if [ -n "${TEMPEST_EXPECTED_FAILURES_LIST}" ] && [ -s ${TEMPEST_LOGS_DIR}stestr_failing.txt ] && [ -s "${TEMPEST_EXPECTED_FAILURES_LIST}" ]; then
     echo "Failing tests marked as expected failures"
-    grep -Fxf ${TEMPEST_EXPECTED_FAILURES_LIST} ${TEMPEST_LOGS_DIR}stestr_failing.txt
-    if ! grep -Fxv -q -f ${TEMPEST_EXPECTED_FAILURES_LIST} ${TEMPEST_LOGS_DIR}stestr_failing.txt ; then
+    grep -Fxf "${TEMPEST_EXPECTED_FAILURES_LIST}" ${TEMPEST_LOGS_DIR}stestr_failing.txt
+    if ! grep -Fxv -q -f "${TEMPEST_EXPECTED_FAILURES_LIST}" ${TEMPEST_LOGS_DIR}stestr_failing.txt ; then
         RETURN_VALUE=0
     fi
 fi


hjensas:
recheck

hjensas:
```
Errors during downloading metadata for repository 'highavailability':
  - Downloading successful, but checksum doesn't match. Calculated: 92b24d792af3ab787ffb6ec16fdfa1ea1d723cf9a1fae31322ae9fd118ce92ca03e5342282cd0eb11dab9ed2261caec5a77ac45002de5a4a1d50c93ba6764a24(sha512)  Expected: c49a0e38560ffa3864054d3dd2a68b1888ce47928d8b0557f3b23aa81123e72e24c35f67f030cc73a74b6c1f0845422bfebabc9dd7f9ed47fdbe03fa263c8c64(sha512) 
Error: Failed to download metadata for repo 'highavailability': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```

hjensas:
recheck

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -499,10 +499,10 @@ if [ ${TEMPEST_DEBUG_MODE} == true ]; then
     sleep infinity
 fi
 
-if [ -s ${TEMPEST_LOGS_DIR}stestr_failing.txt ] && [ -s ${TEMPEST_EXPECTED_FAILURES_LIST} ]; then
+if [ -n "${TEMPEST_EXPECTED_FAILURES_LIST}" ] && [ -s ${TEMPEST_LOGS_DIR}stestr_failing.txt ] && [ -s "${TEMPEST_EXPECTED_FAILURES_LIST}" ]; then
```
sdatko, line 502: So, the problem with that variable is that is not defined at all in this file ~ it is expected to come from test-operator as environment variable https://github.com/openstack-k8s-operators/test-operator/blob/main/controllers/tempest_controller.go#L432

I think I would just ensure the default value of some /dev/null – to make it also more clear that this variable is not containing a list/string but a path to file...

hjensas: With this change, if it the variable is not set e.g `[ -n "${var}}" ]` is evaluated to FALSE - it skips the entire block. Which would be the right thing to do when no "flaky" tests are set.

Setting it to `/dev/null` by default is less efficient, because then it needs to do IO to check the file exist and is not empty.

hjensas: > Did you really just raise IO-related efficiency argument in Bash script...? 😆

Yes I did. :laughing:  I am :100: for pervasively optimizing even in shell scripts.

hjensas:
recheck

hjensas:
recheck

hjensas:
recheck

container-images/tcib/base/os/tempest/run_tempest.sh
sdatko: Ok, jokes aside – I would still much more like to have default value defined, instead of adding another condition on variable in place were it may be simply unset. For me, this script should have `nounset` option enabled by default – and I will be aiming for that – hence I am more in favor of solution introduced as part of https://github.com/openstack-k8s-operators/tcib/pull/320

hjensas: Also, are we prepared to immidiatly backport the refactoring in #320 ?

sdatko: So, I was just deeply testing it with my team this morning, we are about to merge and backport ASAP

hjensas: > So, I was just deeply testing it with my team this morning, we are about to merge and backport ASAP

ok, let's leave this open. If yours merge soon, feel free to close this one. If there are delays, I think we should do this one - re-basing your re-factoring should not be that painful ... :)

PR #323, opened 2025-08-21
[tempest] Add some brackets+quotes here and there

sdatko:
Following Google Style Guide for Bash, I am introducing consistent usage of quotes and brackets wherever possible. Some variables, like `TEMPEST_ARGS` would require special rework in follow-up patch, because of how they are being used currently (not-a-fast-forward change).

[1] https://google.github.io/styleguide/shellguide.html


danpawlik (approved):
shellcheck approve always cool

jirimacku (approved):

PR #326, opened 2025-09-04
Bump tox tests to python 3.12

vyzigold:
Upstream master now supports python 3.10 to python 3.12. The
requirements can't be satisfied for 3.9 anymore, so bump the
tox workflow to 3.12.

With the higher version of python, bumps to flake8 and pylint
were needed as well. Pylint added some new checks it seems. Those
that report errors at the moment are disabled.

One line was exactly 80 character long. I shortened it to pass the
<= 79 character length check.


rabi (approved):
/lgtm

PR #329, opened 2025-09-11
[ansibletest] Add a branch for the git repo

elfiesmelfie:
The allows for setting the POD_ANSIBLE_GIT_BRANCH var to specify which branch to check out.

The default behaviour is preserved i.e. clone the repo and use the default branch When the new env var is set, the specified branch is checked out in the cloned repo.


elfiesmelfie:
recheck

PR #333, opened 2025-09-19
Add cinder-wsgi in tcib and reuse it from here

raukadah:
https://review.opendev.org/c/openstack/cinder/+/960831 drops Remove pbr's wsgi_scripts from setup.cfg and recommends to use should instead reference the Python module path for this service, ``cinder.wsgi.api``,

 This pr adds a modified version of pbr generated cinder-wsgi script to support newer and older release.
    
It also removes httpd conf sed configuration as it is already handled on operator side.

Closes: [OSPCIX-1052](https://issues.redhat.com//browse/OSPCIX-1052)

### DIFF
--- a/container-images/kolla/wsgi/cinder-wsgi
+++ b/container-images/kolla/wsgi/cinder-wsgi
@@ -1,0 +1,14 @@
+#! /usr/bin/python3 -sP
+# PBR Generated from 'wsgi_scripts'
+
+import threading
+
+try:
+    from cinder.wsgi.api import application
+except ImportError:
+    from cinder.wsgi.wsgi import initialize_application
+    application = None
+    app_lock = threading.Lock()
+    with app_lock:
+        if application is None:
+            application = initialize_application()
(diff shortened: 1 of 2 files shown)


container-images/kolla/wsgi/cinder-wsgi
```diff
@@ -0,0 +1,53 @@
...
+    print("*" * 80)
+    print("STARTING test server cinder.wsgi.wsgi.initialize_application")
+    url = "http://%s:%d/" % (server.server_name, server.server_port)
+    print("Available at %s" % url)
+    print("DANGER! For testing only, do not use in production")
+    print("*" * 80)
+    sys.stdout.flush()
+
+    server.serve_forever()
+else:
```
stephenfin, line 46: You can remove all of this. This starts a test server which - per the warning - is not to be used in production. I don't see any reason to support it.

stephenfin (changes requested):

karelyatin:
Wouldn't be better to do it on RPM side consistantly like done for octavia/designate[1][2]? as should also work for older releases

[1] https://review.rdoproject.org/r/c/openstack/octavia-distgit/+/57669
[2] https://review.rdoproject.org/r/c/openstack/designate-distgit/+/57670

fmount (approved):
/lgtm

karelyatin:
> > Wouldn't be better to do it on RPM side consistantly like done for octavia/designate[1][2]? as should also work for older releases
> > [1] https://review.rdoproject.org/r/c/openstack/octavia-distgit/+/57669 [2] https://review.rdoproject.org/r/c/openstack/designate-distgit/+/57670
> 
> That sounds also good. In order to unblock the current ci, I am going with this one. We will do a follow up on rdo side.

ack it was more of an approach to consider, if this is agreed upon followup is fine.
This issue got leaked as cinder spec added wildcards[1], i think that should also be fixed

[1] https://github.com/rdo-packages/cinder-distgit/blob/rpm-master/openstack-cinder.spec#L331

PR #335, opened 2025-09-26
Revert "Add cinder-wsgi in tcib and reuse it from here"

raukadah:
Reverts openstack-k8s-operators/tcib#333

Closes: OSPCIX-1052


abays (approved):
Approving since a SME agrees

/lgtm

PR #364, opened 2026-01-26
Remove override for load_balancer.test_server_path

gthiemonge:
This override prevents the admin/user from overrriding the path of the test_server application. As we are removing the binary from the tempest container to pull it dynamically from tempestconf, it becomes necessary to be able to set a writeable path for non-root users.

Depends-On: https://review.opendev.org/c/openinfra/python-tempestconf/+/973774

JIRA: [OSPRH-25530](https://issues.redhat.com//browse/OSPRH-25530)


kstrenkova (approved):

stuggi (approved):
/lgtm

PR #370, opened 2026-02-17
pin setuptools < 82 before installing the pip packages

lmiccini:
This fixes the ModuleNotFoundError for pkg_resources which was removed in setuptools 82+. The fix ensures setuptools is pinned to version < 82 in two places:

1. In the pre-commit config for pylint to fix CI failures
2. In the horizontest container build for runtime environment


fmount:
lgtm thanks @lmiccini

karelyatin (approved):

PR #373, opened 2026-03-12
[tcib] horizontest: disable pip build isolation to fix XStatic build …

sauragar:
…failure

The horizontest container build was failing during pip installation with:

ModuleNotFoundError: No module named 'pkg_resources'
ERROR: Failed to build 'XStatic-jQuery'

This occurs because pip build isolation creates a temporary build environment that does not use the pre-installed setuptools package. Some Horizon dependencies (e.g. XStatic-* packages) rely on pkg_resources, which is provided by setuptools, causing the build to fail.

Disable pip build isolation so that the build uses the existing environment where setuptools is already installed.

This ensures Horizon dependencies build correctly during the TCIB container image build.

### DIFF
--- a/container-images/tcib/base/os/horizontest/horizontest.yaml
+++ b/container-images/tcib/base/os/horizontest/horizontest.yaml
@@ -1,12 +1,12 @@
 tcib_envs:
   USE_EXTERNAL_FILES: true
-  PIP_BUILD_CONSTRAINT: /tmp/horizon-build-constraints.txt
 tcib_actions:
 - run: bash /usr/local/bin/uid_gid_manage {{ tcib_user }}
 - run: dnf -y install {{ tcib_packages.common | join(' ') }} && dnf clean all && rm -rf /var/cache/dnf
 - run: python3 -m pip install --upgrade pip 'setuptools<82'
 - run: echo "setuptools<82" > /tmp/horizon-build-constraints.txt
-- run: pip install {{ tcib_packages.pip_packages | join(' ') }}
+- run: pip install --no-build-isolation -c /tmp/horizon-build-constraints.txt horizon
+- run: pip install -c /tmp/horizon-build-constraints.txt {{ tcib_packages.pip_packages | join(' ') }}
 - run: cp /usr/share/tcib/container-images/tcib/base/os/horizontest/horizontest_sudoers /etc/sudoers.d/horizontest_sudoers
 - run: chmod 440 /etc/sudoers.d/horizontest_sudoers
 - run: mkdir -p /var/lib/horizontest/external_files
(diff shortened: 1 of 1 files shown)


sauragar:
recheck

container-images/tcib/base/os/horizontest/horizontest.yaml
```diff
@@ -1,12 +1,12 @@
...
   USE_EXTERNAL_FILES: true
-  PIP_BUILD_CONSTRAINT: /tmp/horizon-build-constraints.txt
 tcib_actions:
 - run: bash /usr/local/bin/uid_gid_manage {{ tcib_user }}
 - run: dnf -y install {{ tcib_packages.common | join(' ') }} && dnf clean all && rm -rf /var/cache/dnf
 - run: python3 -m pip install --upgrade pip 'setuptools<82'
 - run: echo "setuptools<82" > /tmp/horizon-build-constraints.txt
-- run: pip install {{ tcib_packages.pip_packages | join(' ') }}
+- run: pip install --no-build-isolation -c /tmp/horizon-build-constraints.txt horizon
+- run: pip install -c /tmp/horizon-build-constraints.txt {{ tcib_packages.pip_packages | join(' ') }}
```
kstrenkova, line 9: Why can't we do `--no-build-isolation -c /tmp/horizon-build-constraints.txt` here, but instead we do special treatment for horizon?

sauragar: **The Problem:**                                                                                                                                                                                                                               
Horizon has dependencies (XStatic-* packages) that require pkg_resources from setuptools during their build process. When pip uses build isolation (the default), it creates a temporary, clean build environment that doesn't include pre-installed packages. This causes the build to fail with:                                                            
```
  ModuleNotFoundError: No module named 'pkg_resources'
  ERROR: Failed to build 'XStatic-jQuery'       
```                                                                         
                                                                                                                         
**Why Not Use --no-build-isolation for Everything?**                                                                                                                                                                                  
Build isolation is a security and reproducibility best practice. Disabling it should only be done when necessary because:                                                                                                                       
  1. Security: Build isolation prevents build scripts from accessing or modifying the system environment                 
  2. Reproducibility: Ensures packages are built in a clean, controlled environment
  3. Dependency correctness: Forces packages to declare their build dependencies properly                                
                                                                                                                         
The other packages (pytest, Django, selenium, etc.) don't have this issue and should be installed with build isolation enabled.                                                                                                               
                                                                                                                         
**The Solution:** 
Install horizon separately with --no-build-isolation (line 8) while keeping build isolation enabled for all other packages (line 9). This minimizes the security/reproducibility risk by only disabling isolation where absolutely necessary.

karelyatin (changes requested):
This PR is not needed, actual issue is periodic pipeline specific as not yet have the last fix https://github.com/openstack-k8s-operators/tcib/pull/368

sauragar:
Thanks @karelyatin for confirming, I am closing this PR.

PR #371, opened 2026-03-03
Use rabbitmq rpms from centos stream Messaging SIG repo

lmiccini:
This commit switches the rabbitmq container image to install rpms coming from the Messaging SIG repo (rabbitmq 4.x).

PR #378, opened 2026-04-01
[antelope] Use rabbitmq rpms from centos stream Messaging SIG repo

lmiccini:
- Switches the rabbitmq container image to install rpms from the Messaging SIG repo (rabbitmq 4.x)
- Unlike #371 (main branch), the antelope base image already has the `extras-common` repo configured, so we enable it directly instead of using `--repofrompath`, which causes a "Repository extras-common is listed more than once" error
- Supersedes #377


abays (approved):
/lgtm

PR #381, opened 2026-04-02
Revert "[antelope] Use rabbitmq rpms from centos stream Messaging SIG repo"

lmiccini:
Reverts openstack-k8s-operators/tcib#378


karelyatin (approved):
/lgtm
We should avoid using repo setup within container files as this can cause issues when pulled downstream
Also there are ways to include tasks based on distro but still good to manage repos outside of containerfiles

PR #391, opened 2026-05-12
[tobiko] Remove quotes for TOBIKO_TESTENV variable

kstrenkova:
Recently a change added quotes to Tobiko variables to follow best practices. However, the environment variable quotation prevented using "scenario -- path/to/test.py" tox syntax, where word splitting is required to pass arguments after --. This change removes quotes to enable this use case.


eduolivares (approved):

abays (approved):
/lgtm

PR #398, opened 2026-06-17
Install frr in neutron-ovn-agent

mtomaska:
Neutron ovn agent evpn extension needs access to `vtysh` command. Unfortunately, `vtysh` does not come as a standalone client and must be installed as part of frr package.

JIRA:  https://redhat.atlassian.net/browse/OSPRH-31097


karelyatin (approved):
/lgtm

PR #404, opened 2026-07-03
Make aarch64 boot asset download optional with graceful fallback <JIRA: OSPRH-32103>

mumesan:
**Description:**
The `extract-boot-assets.sh` script runs and fails when aarch64 packages are not found in x86_64 repositories. This PR makes the download section graceful, there is a check for whether the dnf download errors out and prints a warning log if so. The aarch64 directories are only created when the dnf download succeeds so that the `pxe-init.sh` script in ironic-operator does not run into issues when it looks for those subdirectories and they are not found.

Jira Link: [OSPRH-32103](https://redhat.atlassian.net/browse/OSPRH-32103)


juliakreger (approved):
LGTM, thanks!

PR #408, opened 2026-07-22
Fix horizontest admin auth by loading password from OS cloud config

HanzJas:
The pod injects ADMIN_PASSWORD=12345678, which does not match the RHOSO admin credentials used by --os-cloud default after switch to passwords and secrets dynamic generation. Read the password via OpenStackConfig so horizon.conf matches the working CLI auth.

### DIFF
--- a/container-images/tcib/base/os/horizontest/run_horizontest.sh
+++ b/container-images/tcib/base/os/horizontest/run_horizontest.sh
@@ -39,8 +39,12 @@ SERVICES_REGION_BTN_TEXT_PATTERN="Managing Region {region}"
 SERVICES_REGION_DROPDOWN_XPATH=".//li[@id='services_region_switcher']"
 
 # assert mandatory variables have been set
-[[ -z ${ADMIN_USERNAME} ]] && echo "ADMIN_USERNAME not set" && exit 1
-[[ -z ${ADMIN_PASSWORD} ]] && echo "ADMIN_PASSWORD not set" && exit 1
+if [[ -z "${ADMIN_USERNAME}" ]]; then
+    ADMIN_USERNAME=$(python3 -c "from openstack.config import OpenStackConfig; print(OpenStackConfig().get_one(cloud='default').auth['username'])")
+fi
+if [[ -z "${ADMIN_PASSWORD}" ]]; then
+    ADMIN_PASSWORD=$(python3 -c "from openstack.config import OpenStackConfig; print(OpenStackConfig().get_one(cloud='default').auth['password'])")
+fi
 [[ -z ${DASHBOARD_URL} ]] && echo "DASHBOARD_URL not set" && exit 1
 [[ -z ${AUTH_URL} ]] && echo "AUTH_URL not set" && exit 1
 [[ -z ${REPO_URL} ]] && REPO_URL="https://review.opendev.org/openstack/horizon"


container-images/tcib/base/os/horizontest/run_horizontest.sh
```diff
@@ -38,6 +38,8 @@ SERVICES_REGION_NAME_XPATH=".//li[@id='services_region_switcher']/a"
 SERVICES_REGION_BTN_TEXT_PATTERN="Managing Region {region}"
 SERVICES_REGION_DROPDOWN_XPATH=".//li[@id='services_region_switcher']"
 
+ADMIN_PASSWORD=$(python3 -c "from openstack.config import OpenStackConfig; print(OpenStackConfig().get_one(cloud='default').auth['password'])")
```
kstrenkova: Hi 👋 I think this approach will work, however, I would ask you to give me some time to try to implement it in test-operator. TCIB is the lowest level of this chain and maybe having it in test-operator would be a bit clearer. Also, this means the `ADMIN_PASSWORD` can't be overwritten right? If we did it in test-operator, we can set it as a default, but still give users an option to overwrite it.

HanzJas: Hello @kstrenkova ,
Thank you, good points!

1. I implemented it this way, in horizontest TCIB because it is probably (as far as I know) problem only in horizontests. So fix in test-operator can be maybe overkill? But maybe it is a more clear approach or I am missing something. I am really not sure here so I appreciate you jumping in.

1. The overwriting is another interesting topic. I went through the jobs and I found,
    In component jobs:
```
    cifmw_test_operator_horizontest_repo_url
    cifmw_test_operator_horizontest_horizon_repo_branch
    cifmw_test_operator_horizontest_dashboard_url
    cifmw_test_operator_horizontest_auth_url
    Cifmw_test_operator_horizontest_extra_flag
```
   In operator job:
```
   cifmw_test_operator_horizontest_auth_url
   cifmw_test_operator_horizontest_dashboard_url
   cifmw_test_operator_horizontest_extra_flag
   cifmw_test_operator_horizontest_project_name_xpath
   cifmw_test_operator_horizontest_extra_mounts
```
   In uni jobs:
```
   cifmw_test_operator_horizontest_registry
   cifmw_test_operator_horizontest_namespace
   cifmw_test_operator_horizontest_container
   cifmw_test_operator_horizontest_image_tag
   cifmw_test_operator_horizontest_repo_url
   cifmw_test_operator_horizontest_horizon_repo_branch
   cifmw_test_operator_horizontest_extra_flag
```
I am not saying I did not overlook anything. But it seems like we do not overwrite admin password in any job and I do not have any idea why we would need to do it. Maybe we do not even need to have this (and multiple other) variables as overwritable?
 
But if you have a better/more clear solution, let's do it.

kstrenkova: I agree with you, but test-operator is shipped to customers, so we can't just take parameters away. Your change would make it so the parameter can't be set in CR without it then being overwritten here in this script.
What I am suggesting is to prepare to deprecate both `AdminPassword` and `AdminUsername` in test-operator and move to getting these values from clouds.yaml and  secure.yaml. Something like code below would set it correctly from OpenstackConfig, but also let users set it in CR:
```
if [[ -z "${ADMIN_USERNAME}" ]]; then
      ADMIN_USERNAME=$(python3 -c "from openstack.config import OpenStackConfig; print(OpenStackConfig().get_one(cloud='default').auth['username'])")
fi
if [[ -z "${ADMIN_PASSWORD}" ]]; then
      ADMIN_PASSWORD=$(python3 -c "from openstack.config import OpenStackConfig; print(OpenStackConfig().get_one(cloud='default').auth['password'])")
fi
```
In the future once the deprecation notice is old enough, we can remove this conditional and the parameters safely.

Edit: I already have PRs [1][2] that should make sure there is no default for the `AdminPassword` and `AdminUsername` parameters, so this fix could work. (But it doesn't have to wait for them, as it is currently broken anyway, unless you have some testing in mind)
[1] https://github.com/openstack-k8s-operators/test-operator/pull/473
[2] https://github.com/openstack-k8s-operators/ci-framework/pull/4071

xtmprsqzntwlfb (approved):
The code looks good to me, thank you.

container-images/tcib/base/os/horizontest/run_horizontest.sh
HanzJas: Okay, thanks, let’s wait for your changes to be merged so I can properly test this new code (which should allow us also set user/password from CR).

container-images/tcib/base/os/horizontest/run_horizontest.sh
```diff
@@ -39,8 +39,12 @@ SERVICES_REGION_BTN_TEXT_PATTERN="Managing Region {region}"
 SERVICES_REGION_DROPDOWN_XPATH=".//li[@id='services_region_switcher']"
 
 # assert mandatory variables have been set
-[[ -z ${ADMIN_USERNAME} ]] && echo "ADMIN_USERNAME not set" && exit 1
-[[ -z ${ADMIN_PASSWORD} ]] && echo "ADMIN_PASSWORD not set" && exit 1
+if [[ -z "${ADMIN_USERNAME}" ]]; then
+      ADMIN_USERNAME=$(python3 -c "from openstack.config import OpenStackConfig; print(OpenStackConfig().get_one(cloud='default').auth['username'])")
```
kstrenkova: Nitpick: Could you please make it 4-space indentation to keep it consistent?

kstrenkova (commented):
I have tested this change locally and it appears to be working correctly. I also believe it is the right step logically.

Please don't forget to also cherry-pick this change to antelope once the one nitpick is resolved :D

kstrenkova (approved):

abays (approved):
/lgtm

PR #412, opened 2026-08-03
[tempest] Fix phantom test results from timing-data archive

imatza-rh:
When `discover-tempest-config` fails and `tempest run` never executes,
`stestr last --subunit` picks up old test results from the timing-data
archive and reports them as if they passed in this run.

## Root Cause

The [timing-data archive](https://gitlab.cee.redhat.com/ci-framework/timing-data)
contains the full `.stestr/` directory from previous runs — including old test
result files — not just the timing database that stestr uses for test scheduling.
When `discover-tempest-config` fails, those old results are picked up by the
unconditional `generate_test_results` call.

Example: https://sf.apps.int.gpc.ocp-hub.prod.psi.redhat.com/logs/25f/components-integration/25ffb2f08d194ccba0123053d8dfb21b/

## Fix

Two changes in `run_tempest.sh`:

1. **Extract only timing database files** from the archive (`times.dbm*`)
   instead of the entire `.stestr/` directory. stestr uses `dbm.dumb` which
   stores timing data as three files: `.bak`, `.dat`, `.dir`.

2. **Guard result collection** — track whether `tempest run` actually executed
   and skip `move_tempest_log` + `generate_test_results` when it did not.

Both layers are complementary: the tar filter prevents old results from being
extracted, and the execution guard prevents result collection when tests did
not run.

## Testing

- Verified `stestr` source: uses `from dbm import dumb as my_dbm` for `times.dbm`
- Verified archive structure matches `save_config_files` output
- Tested tar wildcard extraction locally (only `times.dbm.*` files extracted)
- Verified `RETURN_VALUE` semantics are preserved (3 cases: both succeed, discover fails, tempest fails)
- Verified bash ERR trap interaction: `if` condition suppresses ERR trap (matching original `&&` behavior)
- Verified `rerun_failed_tests` and `check_expected_failures` handle missing `FAILED_TESTS_FILE` correctly
- bashate passes clean

Closes: https://redhat.atlassian.net/browse/OSPNW-1674

### DIFF
--- a/container-images/tcib/base/os/tempest/run_tempest.sh
+++ b/container-images/tcib/base/os/tempest/run_tempest.sh
@@ -89,6 +89,7 @@
 
 set -x
 RETURN_VALUE=0
+TEMPEST_TESTS_RAN=false
 
 HOMEDIR=/var/lib/tempest
 TEMPEST_PATH="${HOMEDIR}/"
@@ -344,14 +345,19 @@ function run_tempest {
     upload_extra_images
 
     if [[ -n "${TEMPEST_TIMING_DATA_URL:-}" ]]; then
-        curl -L "${TEMPEST_TIMING_DATA_URL}" | tar -xz --strip-components=1 -C "${TEMPEST_DIR}/.stestr"
+        curl -L "${TEMPEST_TIMING_DATA_URL}" | tar -xz --strip-components=1 -C "${TEMPEST_DIR}/.stestr" --wildcards '*/times.dbm*'
     fi
 
     mkdir -p "${TEMPEST_LOGS_DIR}"
 
-    discover_tempest_config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
-    && tempest run ${TEMPEST_ARGS}
-    RETURN_VALUE=$?
+    if discover_tempest_config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES}; then
+        TEMPEST_TESTS_RAN=true
+        tempest run ${TEMPEST_ARGS}
+        RETURN_VALUE=$?
+    else
+        RETURN_VALUE=$?
+        echo "discover-tempest-config failed (rc=${RETURN_VALUE}), skipping tempest run"
+    fi
 
     run_tempest_cleanup
 
@@ -574,11 +580,15 @@ fi
 
 print_config_files
 save_config_files
-move_tempest_log tempest_results.log
-generate_test_results tempest_results
 
-rerun_failed_tests
-check_expected_failures
+if [ "${TEMPEST_TESTS_RAN}" = true ]; then
+    move_tempest_log tempest_results.log
+    generate_test_results tempest_results
+    rerun_failed_tests
+    check_expected_failures
+else
+    echo "Tests did not run, skipping result collection"
+fi
 
 # Keep pod in running state when in debug mode
 if [ "${TEMPEST_DEBUG_MODE}" == true ]; then


container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -344,14 +345,19 @@ function run_tempest {
...
 
     mkdir -p "${TEMPEST_LOGS_DIR}"
 
-    discover_tempest_config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES} \
-    && tempest run ${TEMPEST_ARGS}
-    RETURN_VALUE=$?
+    if discover_tempest_config ${TEMPESTCONF_ARGS} ${TEMPESTCONF_OVERRIDES}; then
+        TEMPEST_TESTS_RAN=true
+        tempest run ${TEMPEST_ARGS}
+        RETURN_VALUE=$?
```
kstrenkova, line 356: Maybe it is not necessary to do the `if else` and make it a bit less complicated? Like so:
```suggestion
fi

RETURN_VALUE=$?
```

container-images/tcib/base/os/tempest/run_tempest.sh
```diff
@@ -574,8 +580,13 @@ fi
 
 print_config_files
 save_config_files
-move_tempest_log tempest_results.log
-generate_test_results tempest_results
+
+if [ "${TEMPEST_TESTS_RAN}" = true ]; then
```
kstrenkova, line 584: I think starting from `print_config_files` and ending with `check_expected_failures`, everything can be skipped if no tests ran 🤔 But that begs the question why can't we just exit early (although debug mode is a problem). I am open to discuss this part.

container-images/tcib/base/os/tempest/run_tempest.sh
imatza-rh: Moving `RETURN_VALUE=$?` after `fi` would lose the failure exit code - in bash, `$?` after `fi` is 0 when the condition is false (regardless of what the condition returned). Quick demo:

```bash
f() { return 42; }; if f; then echo x; fi; echo $?
# outputs: 0
```

So if `discover_tempest_config` fails and we capture `$?` after `fi`, the script exits 0 (SUCCESS) instead of reflecting the failure. The `if/else` is needed to capture the correct exit code in both branches.

container-images/tcib/base/os/tempest/run_tempest.sh
imatza-rh: Good point - moved `rerun_failed_tests` and `check_expected_failures` inside the `TEMPEST_TESTS_RAN` guard. They already self-guard (`rerun_failed_tests` checks `RERUN_FAILED_TESTS` flag and `FAILED_TESTS_FILE` existence, `check_expected_failures` checks `FAILED_TESTS_FILE`), but wrapping them makes the intent explicit.

Kept `print_config_files` and `save_config_files` outside - they show include/exclude lists and archive configs, which is useful for debugging when `discover_tempest_config` fails.

Re: early exit - it would need to duplicate the debug mode `sleep infinity` at the end of the script, so the current guard approach is simpler.

container-images/tcib/base/os/tempest/run_tempest.sh
kstrenkova: Hmm, alright. Then we can leave it as it is. I just liked the simple approach with the one-liner before, but if we want to set the boolean I guess there is no better solution 😄

container-images/tcib/base/os/tempest/run_tempest.sh
kstrenkova: Ack, thanks for moving the functions in.  I do feel this version is the best we could do here.

kstrenkova (approved):
The code itself looks good to me. Normally, I would also test it locally when I am reviewing a patch. However, due to having a load of work assigned at the moment, I will believe you tested the functionality 😄

abays (approved):
/lgtm

